PR #12, opened 2016-06-29
docs: Fix a typo

jktjkt:
So that the example can be copy-pasted and it builds again.


rollbear:
Thanks. That was a bit sloppy of me.

PR #26, opened 2017-01-26
Make sequence moveable.

mlimber:
This patch makes `trompeloeil::sequence` (and its dependencies) moveable so that we can use "[almost always auto](https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/)" style:
```cpp
auto seq = trompeloeil::sequence{};
```
Still testing it, but I wanted to get your feedback on whether it is worth doing?

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -1,0 +1,1 @@
+theme: jekyll-theme-modernist
--- a/trompeloeil.hpp
+++ b/trompeloeil.hpp
@@ -822,27 +822,43 @@ namespace trompeloeil
   class list_elem
   {
   public:
+    // Moveable but not copyable
     list_elem(list_elem const&) = delete;
     list_elem& operator=(list_elem const&) = delete;
+
     list_elem(
       list_elem &&r)
     noexcept
-      : next(r.next)
-      , prev(&r)
     {
-      r.invariant_check();
-
-      next->prev = this;
-      r.next = this;
-
-      TROMPELOEIL_ASSERT(next->prev == this);
-      TROMPELOEIL_ASSERT(prev->next == this);
-
-      r.unlink();
-
-      TROMPELOEIL_ASSERT(!r.is_linked());
-      invariant_check();
+      *this = std::move(r);
     }
+
+    list_elem& 
+    operator=(
+      list_elem &&r)
+    noexcept
+    {
+      if( this != &r )
+      {
+        next = r.next;
+        prev = &r;
+
+        r.invariant_check();
+
+        next->prev = this;
+        r.next = this;
+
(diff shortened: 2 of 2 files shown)


rollbear:
I like the idea very much, but I'm not quite happy with your suggested solution.

I did a quick hack which on a superficial glance seems to work:

```diff
diff --git a/compiling_tests.cpp b/compiling_tests.cpp
index 0cbc1dc..d66aa21 100644
--- a/compiling_tests.cpp
+++ b/compiling_tests.cpp
@@ -159,7 +159,7 @@ TEST_CASE_METHOD(Fixture, "follow single sequence gives no reports", "[sequences
   {
     mock_c obj1(1), obj2("apa");
 
-    trompeloeil::sequence seq;
+    auto seq = trompeloeil::sequence{};
 
     REQUIRE_CALL(obj1, count())
       .IN_SEQUENCE(seq)
diff --git a/trompeloeil.hpp b/trompeloeil.hpp
index 3cedcf1..63e48f0 100644
--- a/trompeloeil.hpp
+++ b/trompeloeil.hpp
@@ -934,6 +934,8 @@ namespace trompeloeil
   class list : private list_elem<T>, private Disposer
   {
   public:
+    list() noexcept;
+    list(list&&) noexcept;
     ~list();
     class iterator;
     iterator begin() const noexcept;
@@ -1016,6 +1018,10 @@ namespace trompeloeil
     list_elem<T>* p;
   };
 
+
+  template <typename T, typename Disposer>
+  list<T, Disposer>::list() noexcept = default;
+
   template <typename T, typename Disposer>
   list<T, Disposer>::~list()
   {
@@ -1028,6 +1034,9 @@ namespace trompeloeil
   }
 
   template <typename T, typename Disposer>
+  list<T, Disposer>::list(list&&) noexcept = default;
+
+  template <typename T, typename Disposer>
   auto
   list<T, Disposer>::begin()
   const
@@ -1085,7 +1094,7 @@ namespace trompeloeil
   {
   public:
     sequence() noexcept = default;
-    sequence(sequence&&) = delete;
+    sequence(sequence&&) = default;
     ~sequence();
 
     bool
```

Do you think this is good enough? I have not checked on VisualStudio (which is always needed, it seems to differ quite a lot from g++ and clang++.) If this works satisfactorily for you, and you can verify that it's OK on VisualStudio too, please add more cases to the `compiling_tests.cpp` to ensure it is (and stays)  alright and then I'd gladly accept a pull request.

rollbear:
Also, can you please work from branch `develop` instead? It's not wildly experimental, and not hugely different from `master`. I just like to let things mature there for a short while before going public.

mlimber:
Moved to develop.

PR #27, opened 2017-01-26
Make trompeloeil::sequence moveable

mlimber:
This patch makes `trompeloeil::sequence` (and its dependencies) moveable so that we can use "[almost always auto](https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/)" style:
```cpp
auto seq = trompeloeil::sequence{};
```
I have moved to the develop branch, per the pull request on the master branch [here](https://github.com/rollbear/trompeloeil/pull/26).

Note that, per the [Rule of 5](http://stackoverflow.com/a/4782927/201787), it should have move assignment also, which I've added. (Here, we're relying on implicit deletion of the copy operations, though the [CppCoreGuidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c21-if-you-define-or-delete-any-default-operation-define-or-delete-them-all) are slightly more strict and say one shouldn't rely in implicit generation/deletion at all.)

I did build and run on Visual Studio 2015 Update 3, but it reports the following on your develop branch:
```
tests.exe is a Catch v1.6.0 host application.
Run with -? for options

matching calls are traced
compiling_tests.cpp(4009)

compiling_tests.cpp(4055): FAILED:
  REQUIRE( std::regex_search(os.str(), std::regex(re)) )
due to unexpected exception with message:
  compiling_tests.cpp:4015
  obj1.getter(_, _) with.
    param  _1 == 3
    param  _2 == foo

  compiling_tests.cpp:4016
  obj2.foo("bar") with.
    param  _1 == bar
  threw exception: what() = nonono

  compiling_tests.cpp:4018
  obj1.getter(ANY(int)) with.
    param  _1 == 4
   -> 5

  compiling_tests.cpp:4020
  obj2.foo("baz") with.
    param  _1 == baz
  threw unknown exception

  regex_error(error_complexity): The complexity of an attempted match against a
  regular expression exceeded a pre-set level.

test cases: 233 | 232 passed | 1 failed
assertions: 351 | 350 passed | 1 failed
```
It reports the same with my changes (it would fail to build if the changes were bad). This could be an instance of [this "problem"](https://connect.microsoft.com/VisualStudio/feedback/details/750453/stl-regex-poor-performance-in-debug-builds-crash-in-release-builds).

Everything passes with GCC 5.4 and Clang 3.8.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -15,6 +15,8 @@
 	    ...
 	  }
 
+	* Made sequence and its dependencies moveable. (mlimber)
+
 v22 2016-12-13
 
 	* Messages from violations of expectations now print collections
--- a/compiling_tests.cpp
+++ b/compiling_tests.cpp
@@ -159,7 +159,7 @@ TEST_CASE_METHOD(Fixture, "follow single sequence gives no reports", "[sequences
   {
     mock_c obj1(1), obj2("apa");
 
-    trompeloeil::sequence seq;
+    auto seq = trompeloeil::sequence{};
 
     REQUIRE_CALL(obj1, count())
       .IN_SEQUENCE(seq)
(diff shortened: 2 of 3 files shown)


rollbear:
Thanks. Looks good. I haven't seen the Visual Studio regex error before, which is a bit strange.

I may add some more tests to the list<> during the weekend, just to make sure.

PR #30, opened 2017-02-25
Add dash as in header-only, thread-safe

martinmoene:
hmm, reissued for develop branch.

Also, update repository's description at the top.


martinmoene:
hmm, reissued for develop branch.

PR #31, opened 2017-02-25
Add dash as in header-only, thread-safe

martinmoene:
Also, update repository's description at the top.


rollbear:
Thanks @martinmoene

PR #47, opened 2017-07-22
Develop

apmccartney:
Well that was a doozy. Here's a summary of my modification:

+ Moved to Trusty for better support and enabled sudo
+ Clang builds also download GCC-6 to have access to C++14 headers
+ Problems related to ASAN in Clang were the result of improper software packaging. These runs have an extended pre-script step which resolves those issues. This required sudo access.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ matrix:
         apt:
           sources:
             - ubuntu-toolchain-r-test
-            - george-edison55/cmake-3.x
+            - ppa:george-edison55/cmake-3.x
           packages:
             - g++-4.9
             - libcurl4-openssl-dev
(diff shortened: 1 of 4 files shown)


rollbear:
Wow! Huge thanks.

What is the reason that clang-3.7 and clang-3.9 are fetched as .tar.gz which overwrites the installed clang? I want to understand this before pressing the merge button.

apmccartney:
No worries!

You're correct. In both 3.7 and 3.9, we're downloading a full installation but not actually overwriting the system clang executables acquired from apt. 

In the clang 3.7 run, the apt package was missing the clang runtime library providing the address sanitizer. By using `cp -n`, we copy any libraries in installation extracted from the tar that were missing in the apt installation. 

In the clang 3.9 run, the ASAN library existed but seemed to required a version of gold greater than was available through the package manager or any public ppa I could find. In this case we overwrote the clang libraries distributed with the apt package with the llvm tar.

rollbear:
Added comments describing the work arounds. Thanks again.

PR #49, opened 2017-07-28
Fix issue 48: undefined reference with clang++-3.5

AndrewPaxie:


AndrewPaxie:
Not sure what to do about the AppVeyor failure.

rollbear:
Never mind the appveyor error. I enabled appveyor a few days ago, thinking basic support would be easy. Many errors later I removed the `appveyor.yml` file, and that's the result you're seeing. I just changed the appveyor settings to only build a (currently nonexisting) branch `appveyor`. That'll hopefully silence it until I have a working appveyor setup.

On the patch: `std::move(re)` should be used here, rather than `std::forward<std::regex>(re)`, should it not? `re` is explicitly an rvalue reference and not a forwarding reference in a function template.

AndrewPaxie:
You are of course correct, thanks to catching my mistake.

When such things happen, I reminds me that I am

(a) Forgetful.
(b) Optimistic.

I normally think of `&&` as a forwarding reference and forgot the
rvalue reference case, which is obviously what `regex_check` uses.
I have reviewed Scott Meyers' items on this topic and the relevant
articles on cppreference.com.

My optimism came from the hope that thinking mistakes would be
caught by a tool sometime before a reviewer's time was taken up
with my pull request.  I reviewed carefully my development process
and found that

- CLion (2017.2) had nothing to say about `std::forward` being applied
  to an explicit rvalue reference.
- None of the compilers I used (loosely: clang 3.5-4.0, g++ 4.8-7.1, VS 2017)
  reported a warning.
  
Missing from this process were static analysis tools.  I wondered
if any of the ones I was familiar with could have helped, had I used them.

The answer is 'no', and I tried the following:

- Visual Studio Community 2017 `/analyze`.
- `clang-check` 4.0.0 `-analyze`.
- `cppcheck` 1.76.1 (Zesty default), 1.79, and 1.80 (released today).

While `clang-check` did report four 'Value stored is never read' warnings,
it and the other tools were silent about the `std::forward` error.

Again recalling Meyers, this time from his
["Things that Matter"](https://www.youtube.com/watch?v=RT46MpK39rQ)
talk, given at DConf 2017:

- Never stop learning.
- The "toolability" of C\+\+ is poor.

I will investigate what it takes to add a check for this error
to `clang-check` or `cppcheck`.  I am optimistic, how hard can it be?

rollbear:
In this case it was harmless, since forward became a move, but yes, it's problematic. I think it's one of the biggest mistakes of recent years that the exact same syntax is used for rvalue reference and for forwarding reference. It also additionally makes it quite difficult to write a function that accepts a templated rvalue parameter (and not an lvalue.)

Anyway, thanks for spotting and fixing this. I have been aware of the clang-3.5 link error, but I thought it was due to a library problem and ignored.

PR #53, opened 2017-08-09
Add toggle to enable `override` by default for mocked functions

### LINKED ISSUE #51: Make 'override' a default
It would be nice if `override` was used by default in the `MAKE_MOCK` and `MAKE_CONST_MOCK` macros. You could turn this on with a preprocessor flag, or based on language version (`__cplusplus` I think). Is there a reason this isn't assumed by default? I can't imagine it's possible (or meaningful) to mock non-virtual interfaces since there is no dispatch without it.

rcdailey:
Define a preprocessor definition named `TROMPELOEIL_OVERRIDE_BY_DEFAULT` to
make `override` the default specifier for all mocked functions. When this
is enabled, the actual `spec` parameter of `TROMPELOEIL_MAKE_MOCK_()` is
ignored.

Fixes #51.

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -1056,7 +1056,10 @@ Make a `const` [mock function](#mock_function) named *func_name*. It is a good
 idea for this to implement a pure virtual function from an interface, but
 it is not a requirement. `n` is the number of parameters in *signature*.
 *specifiers* is an optional list which may include attributes or specifiers like
-[`override`](http://en.cppreference.com/w/cpp/language/override).
+[`override`](http://en.cppreference.com/w/cpp/language/override). If the
+preprocessor definition `TROMPELOEIL_OVERRIDE_BY_DEFAULT` is specified prior to
+including `trompeloeil.hpp`, then `override` is always used for every mock
+function. In this case, the `specifiers` argument is ignored.
 
 Example:
 ```Cpp
@@ -1099,7 +1102,10 @@ Make a non-const [mock function](#mock_function) named *func_name*. It is a
 good idea for this to implement a pure virtual function from an interface, but
 it is not a requirement. `n` is the number of parameters in *signature*.
 *specifiers* is an optional list which may include attributes or specifiers like
-[`override`](http://en.cppreference.com/w/cpp/language/override).
+[`override`](http://en.cppreference.com/w/cpp/language/override).If the
+preprocessor definition `TROMPELOEIL_OVERRIDE_BY_DEFAULT` is specified prior to
+including `trompeloeil.hpp`, then `override` is always used for every mock
+function. In this case, the `specifiers` argument is ignored.
 
 Example:
 ```Cpp
--- a/include/trompeloeil.hpp
+++ b/include/trompeloeil.hpp
@@ -3481,6 +3481,16 @@ namespace trompeloeil
 
 #endif
 
+// Define a preprocessor definition named TROMPELOEIL_OVERRIDE_BY_DEFAULT to
+// make `override` the default specifier for all mocked functions. When this
+// is enabled, the actual `spec` parameter of `TROMPELOEIL_MAKE_MOCK_()` is
+// ignored.
+#if defined(TROMPELOEIL_OVERRIDE_BY_DEFAULT)
+#  define TROMPELOEIL_OVERRIDE_SPEC(...) override
+#else
+#  define TROMPELOEIL_OVERRIDE_SPEC(...) __VA_ARGS__
+#endif
+
 #define TROMPELOEIL_MAKE_MOCK_(name, constness, num, sig, spec, ...)           \
   using TROMPELOEIL_LINE_ID(cardinality_match) =                               \
     std::integral_constant<bool, num == ::trompeloeil::param_list<sig>::size>; \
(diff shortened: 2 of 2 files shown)


rollbear:
Sorry for the long delay in responding. I'm on vacation and not always by my computer.

I like this solution. Thanks. Could you add an entry to the FAQ as well?

rcdailey:
One issue with this solution is that it does not work with specifiers other than override. The way the 3rd parameter is documented makes it seem more generic, but going forward this change will basically make the 3rd argument dedicated to override (not that I can think of any other specifiers that would be used here. Maybe `final`, but that doesn't seem useful for a mock class anyway). 

IMHO, I think override should be implied always, without any option to turn it off (this just adds unnecessary complexity). Are there valid cases where a mock function should not override a base class virtual function?

rollbear:
That is true. `final` is plausible, but doesn't really make sense. Other options are `&`, if you have overloads for a function for r-values and l-values, but that too seems rare. What I was thinking of, though, was attributes like `[[noreturn]]`.

Indeed there are use cases for using mocks that do not implement an interface. Truly, I find that I rarely do. One very frequent example (for me) is when implementing a class template that is tested via mocked collaborators. There's no base class anywhere, and forcing me to write one just to satisfy a tool that's supposed to help me, would make me very grumpy. In fact, the requirement of many mocking frameworks to always test virtual functions via base classes was one of the driving factors behind writing Trompeloeil.

If you have 53 minutes to spare, you can watch my presentation from the NDC{Oslo} conference, where after an initial classic introduction, the rest is done without any inheritance. Although, I think you've used Trompeloeil for long enough now to probably not learn much new from the video.

https://www.youtube.com/watch?v=vvQ-kK4coYM

rcdailey:
Thanks for the information. I just wanted to confirm there are use-cases for this. If you do not feel comfortable making `spec` representing only `override` going forward, I'll need to come up with another idea. I'm trying to find a solution that doesn't involve breaking or changing existing behavior if possible, but it's difficult.

I've thought of a new mock macro, something like `MAKE_MOCK_OVERRIDEn`, but this gets nasty because now you have to duplicate those for each `n` macro, including the `_CONST` versions.

Another flaw with my idea (after reading your last response) is that there's no way to turn `override` off in specific situations (for example, maybe in 99% of cases you want the default, but you want to do an occasional `MAKE_MOCK` without using `override`.

Can you think of any modificiations I can make or maybe entirely different approaches?

rollbear:
I'm not sure what's the best way to go. Creating another set of macros could maybe lead to a combinatorial explosion in the future.

I think it's probably OK to have a case where MAKE_MOCKn is *always* `override`, and then lose the ability to add other specifiers. You can always choose to not define `TROMPELOEIL_OVERRIDE_BY_DEFAULT` before include if you need the specifiers, even if that means having to manually add `override`.

What I'm a bit concerned about is that your implementation discards any specifiers. I'd feel better if it gave a compilation error if there were any (and preferably one that is very explicit about what the problem is.) Otherwise people might get surprising behaviour, perhaps causing long debugging sessions to figure out what's going on.

rollbear:
Just wanna make sure we're not misunderstanding one another.

I appreciate your enthusiasm, and in general I really like your pull request. 

If you can add an entry to the FAQ, a refman entry for the new macro, and very specifically, if you can alter the `MAKE_MOCKn` macros such that accidental modifiers cause some kind of compilation error (the more specific the better,) I'll be glad to push the "merge pull request" button.

rcdailey:
Sorry I didn't mean to ignore you, I actually got super busy at work and forgot to come pay a visit here.

I don't mind updating documentation further as you requested, but a compilation error will require some thought. Let me think on it. I would only want a compilation error if the specifier did not equal blank or override. I'll see if I can use `static_assert` trickery for this.

rollbear:
No hurry. I just wanted to make sure we didn't misunderstand each other and that perhaps you were waiting for me.

rcdailey:
I just realized I had seriously forgotten about this PR... surprised you didn't close it :)

I have rebased my branch to remove the unnecessary whitespace changes my tooling previously introduced. I still haven't thought of a good solution to the diagnostic error I need to add. Another issue I thought of is that manually defining `TROMPELOEIL_OVERRIDE_BY_DEFAULT` all over the place will cause this flag to be non-deterministically propagated across header files that include each other, for the purposes of sharing mocks. For example, `MockA.hpp` may do a `#define TROMPELOEIL_OVERRIDE_BY_DEFAULT` and then `MockB.hpp` includes `MockA.hpp` but doesn't want the default override functionality. This will cause `MockB.hpp` to break.

Granted, in my specific usage, `TROMPELOEIL_OVERRIDE_BY_DEFAULT` is provided at the command line for my whole project. This is simple because I never want to define a mock class that isn't overriding base class methods. I don't have the template situation you described earlier.

This propagation of the preprocessor flag could be a big problem depending on how people commonly use your library. Did you have any thoughts on this latest quirk?

Sorry to talk this to death, I promise to eventually get around to doing some coding. I guess I've lost some motivation since originally I assumed this would be a simple change. Now it's kind of blown up into a complex problem with many issues. It's starting to not feel worth it anymore.

rollbear:
I'm thinking of stealing the way to mock interfaces from http://turtle.sourceforge.net/turtle/reference.html#turtle.reference.creation.member_function

I think this is a much preferable technique, if I can pull it off. (I don't really foresee it being difficult, but I may run into surprises.) I'll try to find the time for an experiment this weekend.

rcdailey:
How does this solve the override problem?

rollbear:
If you look at the `develop` branch, and find the file `test/compiling_tests.hpp` and search for the use of `mock_interface`, you may find something interesting. The signature is obtained from the base class, and these *always* override. I have not yet documented this, and I need more tests, but the basics looks nice.

rcdailey:
Closing this since I've started using `mock_interface` and this exceeds the requirements initially covered by this issue. I'm very happy with the new `IMPLEMENT_MOCK()` macros. Thank you!

PR #56, opened 2017-09-17
Fix issue 55: restore warnings for Clang builds

AndrewPaxie:


rollbear:
Thanks Andrew. I should've seen that one.

PR #58, opened 2017-09-27
Bugfix/issue 57

AndrewPaxie:
Attempting to use the auxiliary functions with, for example, 

`static_assert(std::is_member_function_pointer<e_t>::value`

is not enough to remove the warning, hence taking the address of
the appropriate member function overload and ignoring the variable.

An early revision used a macro `TROMPELOEIL_NONAME_PARAM_LIST(num, sig)`
which generated a parameter list containing only type declarations. 
This was avoided by reusing `TROMPELOEIL_PARAMS(num)` and placing a
dummy parameter (`#name` in this case) in the argument list to handle
the leading comma in the replacement list from the macro expansion.

I have stopped short of making this a Clang-only change, but this
pull request may be too much for other compiler users (GCC, MSVC) to pay.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,6 @@
 	* Fix Issue #48: Move "re" in lambda capture list in "regex_check".
 	* Fix Issue #55: Restore warnings for Clang builds.
+	* Fix Issue #57: Avoid uneeded-member-function warning from Clang.
 
 v28 2017-07-24
 
--- a/include/trompeloeil.hpp
+++ b/include/trompeloeil.hpp
@@ -3316,6 +3316,25 @@ namespace trompeloeil
     return ::trompeloeil::neg_matcher<std::decay_t<M>>{std::forward<M>(m)};
   }
 
+  /*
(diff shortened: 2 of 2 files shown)


AndrewPaxie:
Sorry there seems to be some changes from `master` in this pull request.  You may or may not want to take it as-is even if the intended changes are otherwise satisfactory.

AndrewPaxie:
MacOS build failed, possibly due to load.
See: https://travis-ci.org/AndrewPaxie/trompeloeil/builds/280302521
for a successful build of the branch and
See: https://blog.travis-ci.com/2017-09-22-macos-update

AndrewPaxie:
Cleaned up commit.

rollbear:
I have mixed feelings about this one. Unnecessary code for shutting up what I *really* consider to be a compiler issue.

That said, clean compiles are important, so I guess we'll have to accept ugly work arounds from time to time.

BTW, sorry for taking so long on this. Last week was a really bad week.

PR #59, opened 2017-09-30
Avoid allowing compiler extensions in Trompeleoil

AndrewPaxie:
When compiling with GCC and Clang the C\+\+ dialect selected is `-std=gnu++14`.

This allows compiler extensions to be used in the Trompeloeil code base,
which may be unintended.


rollbear:
Are you sure this is a good thing to add globally for the library? Keep in mind that other cmake projects using Trompeloeil can do so via the cmake package, inheriting these constraints.

I'm not quite sure what's the best way to handle C++ versions. I want people to be able to use extensions if they so please, and of course C++17 too.

Maybe setting the strict C++14 constraint for the test targets only?

I'm also thinking of beginning to add C++17 in trompeloeil.hpp (under `#ifdef`s, because `if constexpr` and fold-expressions are likely to cut compilation time.

AndrewPaxie:
I didn't realise the use case of Trompeloeil included as a cmake package, but now appreciate it is a goal you worked hard to achieve.

I will rework settings, applying them to `compiling_tests` only and submit an updated pull request when done.

Good news on the C++17 start.  I had hoped to complete C++11 before then, but _c'est la vie_.

rollbear:
as an aside, can I tease you with another just started toy project?

https://github.com/rollbear/strong_type

Still in very experimental state. If this is your type of thing, I could use some help with what's a good feature set and naming. I think the idea is sound, but I'm not happy with all the details.

AndrewPaxie:
Well, I have just started reading "Elements of Programming" by Stepanov and McJones, so I may be able to make some comment if not actual commitment in the next few weeks.  I would also like to see how Concepts help achieve (or hinder) your project goals.

AndrewPaxie:
Ready for re-review.

rollbear:
Excellent. Thanks!

PR #61, opened 2017-10-10
Lightly edited documentation

AndrewPaxie:


rollbear:
Wow, that's a massive proofreading/editing effort there. Thanks!

I'm confused about your editing for period/comma together with parenthesis.

I've learned to always (illogically) have the closing parenthesis outside of the period/comma, but you've consistently changed them the other way. Maybe I'm trying to apply a Swedish rule to English, though?

AndrewPaxie:
I can only tell you what I learned as a student about the "parenthesis rules" (for English) and I may have applied them incorrectly:
- If a phrase is in parentheses, finish the phrase and close the parenthesis, then add the full stop _(blah blah blah)_. Same applies to the positioning of a comma _(when necessary)_, which separates clauses.
I think my explanation hints that this rule for parentheses is the opposite of the rule for double quotes, but I won't confuse things further.
- If an entire sentence is in parentheses, the full stop applied to the sentence. _(We don't want any information to escape from here.)_

I think these rules cover all the cases where I felt compelled to make a change.

AndrewPaxie:
Quick and Dirty Tip
http://www.quickanddirtytips.com/education/grammar/periods-and-parentheses

PR #68, opened 2017-12-04
Moved CXX11_AS_STRING to compiling_tests_11.cpp

AndrewPaxie:


rollbear:
Good.

I'm thinking of announcing the `develop` branch as beta-support for C++11 on twitter and G+, where I have a decent enough following for someone to pick it up and try it.

Do you think I should wait longer? Also, do you have accounts on twitter and/or G+, so I can attribute your work properly?

Also, a very minor thing I want to do, at least before going to next tagged stable release, is to compact the ChangeLog. I ususally try to keep it to changes that are visible to users (not library developers,) between releases. Thus changes that reflect self-test or internal refactorings are typically not mentioned (unless they make a noticeable performance difference.) Especially changes that are made and then backed out are not mentioned, because to a library user of tagged releases, the change has never happened.

AndrewPaxie:
This is the right time for a beta support announcement.   Better compatibility between C\+\+ dialects is still the subject of research.

Surprisingly, I have a G+ account: https://plus.google.com/109580206267929675184 .
With regret I am not on Twitter.

Please compact the ChangeLog.  I had no idea how to write one so threw everything in.  it's also not formatted well (is Markdown something to put in a ChangeLog?)  If this means compressing Git history as well, so be it.  I have backups of version one and two of the C++11 API which I will be able to rely on for expository purposes in the future.

rollbear:
Have had a couple of glasses, so I'll delay the announcement until tomorrow. I'll compact the ChangeLog, but leave the git history as is - if anyone is tracking `develop` hell will break loose otherwise.

PR #75, opened 2018-04-02
Updated unit test framework to Catch 2.2.1

AndrewPaxie:


rollbear:
Don't know why I missed this one before tagging v30. Thanks again.

PR #76, opened 2018-04-02
Used markdownlint to make documentation more consistent

AndrewPaxie:
I have started using Visual Studio Code for software development,
mainly because the Markdown Support extension in CLion 2018.1
shipped with critical defects.

One of the extensions I found was `markdownlint`, which is a linter
for Markdown documents.  Details,

Site: <https://github.com/DavidAnson/vscode-markdownlint>
Version: 0.14.1

This pull request is the result of applying `markdownlint` to the
current documentation for Trompeloeil.  Doubtless a few other
improvements to the documentation may be made, but this is
a start.

## Markdownlint configuration

Here is the configuration I used for `markdownlint`, placed in my
User Settings

```json
{
    "markdownlint.config": {
        "MD013": false,
        "MD024": false,
        "MD025": false,
        "MD026": {
            "punctuation": ".,;:!"
        },
        "MD033": {
            "allowed_elements": ["a"]
        }
    }
}
```

Short description of each line,

- MD013: Do not enforce a line length limit.
- MD024: Allow headers with the same content. [Mainly for my blog.]
- MD025: Allow multiple top-level headers. [Mainly for my blog.]
- MD026: Accept question mark as trailing punctuation (for FAQ).
- MD033: Allow HTML anchor in documentation.

This could be refined into a settings file in the `docs/` directory
at a future time.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,5 @@
+	* Used markdownlint to make documentation more consistent
+
 	* Silenced 2 clang++-5 warnings about missing "override"
 
 	* Moved CXX11_AS_STRING macro out of trompeloeil.hpp and into
--- a/docs/Backward.md
+++ b/docs/Backward.md
@@ -1,18 +1,18 @@
-# Backward compatibility with earlier versions of C++
+# Backward compatibility with earlier versions of C\+\+
 
 - [Introduction](#introduction)
-- [C++11 compromises](#cxx11_compromises)
-  - [C++11 API](#cxx11_api)
+- [C\+\+11 compromises](#cxx11_compromises)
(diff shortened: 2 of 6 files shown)


rollbear:
Very neat. Thanks. Do you think it'd be possible to add something like this to `travis-ci`, or `appveyor`? It'd be very good to get a consistency check on the documentation on every push.

AndrewPaxie:
I think there would be a way e.g. invoking an equivalent `node.js` tool from the command line,
but `markdownlint` is a front-end tool.  I will investigate options.

I just fixed > 2300 warnings from my blog notes on the Trompeloeil project.
Reaching zero warnings is a position one wants to maintain, and this is much
easier for me now.

PR #81, opened 2018-05-01
fix c++11 msvc support

Neargye:
Hi,
I have some problems using the trompeloeil in the Visual Studio 2015, it does not [support full С++14](https://msdn.microsoft.com/en-us/library/hh567368.aspx).
Therefore, I think in Visual Studio 2015 need use С++ 11 API trompeloeil.

Maybe it's worth adding  test in Appveyor CI with Visual Studio 2015 and Visual Studio 2017.

### DIFF
--- a/include/trompeloeil.hpp
+++ b/include/trompeloeil.hpp
@@ -74,11 +74,13 @@
 
 #define TROMPELOEIL_GCC_VERSION 0
 
-/* MSVC is an amalgam of C++ versions, with no provision to
- * force C++11 mode.  It also has a __cplusplus macro stuck at 199711L.
- * Assume the C++14 code path.
- */
-#define TROMPELOEIL_CPLUSPLUS 201401L
+#if _MSC_VER > 1910
+#  define TROMPELOEIL_CPLUSPLUS 201401L
+#elif _MSC_VER == 1900
+#  define TROMPELOEIL_CPLUSPLUS 201103L
+#else
+#  error requires Visual Studio 2015 RC or later
+#endif
 
 #endif
 


AndrewPaxie:
Visual Studio 2015 was one configuration I haven't had direct access to test. 
But I would like to know more of the specific C++14 features that fail when using VS 2015.
After all, Trompeloeil is a C++14 library first.

Do you have a specific example of code that fails in VS 2015?

Your suggestion for AppVeyor support is well made.

rollbear:
I agree with @AndrewPaxie.  I don't find anything wrong with the patch, per se (do you @AndrewPaxie?) but would prefer a solution that makes C++14 work with VS 2015.

Neargye:
I need a couple of days to write examples. I will write here as it will be ready.

rollbear:
I do most certainly not mind @andrewpaxie, and please go on.

Neargye:
@AndrewPaxie @rollbear 
I took the original tests, and checked them on msvc with c++14
https://github.com/Neargye/trompeloeil/tree/msvc -> [result](https://ci.appveyor.com/project/Neargye/trompeloeil)
I removed from the cmake everything that does not apply to the tests.
But how can an idea, rewrite tests from scripts to CMake test? Cmake perfectly able to test   С++
compilation_errors [example ](https://github.com/Neargye/trompeloeil/blob/msvc/compilation_errors/CMakeLists.txt)
test [example](https://github.com/Neargye/trompeloeil/blob/msvc/test/CMakeLists.txt)

Neargye:
msvc with c++11
https://github.com/Neargye/trompeloeil/tree/msvc_cpp11 -> [result](https://ci.appveyor.com/project/Neargye/trompeloeil/build/msvc_cpp11%20%231)

AndrewPaxie:
Thanks for creating a run on AppVeyor for review.

The initial error at line 574 of `trompeloeil.hpp` suggests VS2015 is missing `std::conditional_t`,
a detail that `namespace detail` is designed to provide when necessary.

This further suggests a different resolution than this current patch.

I am trialling local builds using `cmake` on Windows before any patches are developed.  A work in progress.

Neargye:
@AndrewPaxie Yes, I agree, the patch helped only in my case.
Can I help you somehow?

AndrewPaxie:
Thank you for the offer.

At the moment I am working with an online compiler
http://rextester.com/l/cpp_online_compiler_visual
that seems to be VS2015.  This is as close to the compiler as I
can get at the moment.

I need to know if `std::conditional` or `std::conditional_t` are in `<type_traits>`
for VS2015.  

I am not sure why the compiler is suggesting `typename` in
front of `std::tuple_size<T>::value` - maybe try that and see what happens!

BTW: the test cases for C++14 are guarded and don't compile in C++11 -
no need to mangle the `CMakeLists.txt` file.

Neargye:
@AndrewPaxie  conditional_t since C++14 http://en.cppreference.com/w/cpp/types/conditional

AndrewPaxie:
Sadly, we cannot rely on cppreference.com when it comes to VS2015, only test programs,
reading the source, and blogs from around the web.  This is the problem of partial C++
support that Microsoft is working to resolve. See e.g.
https://blogs.msdn.microsoft.com/vcblog/2018/04/09/msvc-now-correctly-reports-__cplusplus/ .

Try this test program in your environment:

```cpp
#include <iostream>
#include <type_traits>
#include <tuple>

#if 1

namespace detail {

    using std::conditional_t;

}

#else

namespace detail = std;

#endif

template <size_t N, typename T>
  using conditional_tuple_element
    = detail::conditional_t<(N < std::tuple_size<T>::value),
                         typename std::tuple_element<N, T>::type,
                         int>;

   

typedef std::conditional_t<true, char, long> U;

int main()
{
    std::cout << _MSC_VER;
}
```

When compiled it reproduces the defect.

When `#if 1` changed to `#if 0` then the code compiles, implying a name injection/lookup
defect in VS2015.  This indicates the direction we need to take for restoring VS2015
support.
@rollbear I'm with you on the Compiler Explorer support request.

Neargye:
I think could close this request, it was incorrect and fixed only my case.

Neargye:
@rollbear I have a vs 2015, I'll try to debug this problem.

Neargye:
open issue https://github.com/rollbear/trompeloeil/issues/82

PR #86, opened 2018-05-10
Cleanup ci

Neargye:
I tried to do the same testing on appveyor and travis.


Neargye:
@rollbear Do not close while the requester, I will try to transfer the script completely with check compilation error message.
I will use this particular test for testing.

PR #89, opened 2018-05-22
Improve msvc ci test

Neargye:
Add Debug/Release versions of the build, display warnings as errors on msvc

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -220,9 +220,17 @@ if (MASTER_PROJECT AND CMAKE_BUILD_TYPE MATCHES Debug)
   else()
 
     add_custom_target(
-      run_self_test
+      run_debug_self_test
       COMMAND
         ${CMAKE_CURRENT_BINARY_DIR}/Debug/self_test
+      DEPENDS
+        self_test
+    )
+
+    add_custom_target(
+      run_release_self_test
+      COMMAND
+        ${CMAKE_CURRENT_BINARY_DIR}/Release/self_test
       DEPENDS
         self_test
     )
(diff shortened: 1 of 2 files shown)


Neargye:
It is necessary to correct the cmake a little more

Neargye:
Added a terrible workaround, for the correct run_self_test on msvc.

rollbear:
I'm not super happy with failed builds on `master`, but I think this is exactly what is needed. Am I correct in assuming that this is what built green with @AndrewPaxie's fixes?

Maybe submit this one for `develop` instead? Red on `develop` is OK, and then add the fixes in a separate commit.

Or maybe incorporate the `try`/`catch` work around in this pull request?

I'm undecided.

Neargye:
I'll fix the pool today, and send it to the development branch

AndrewPaxie:
Your "terrible workaround" builds on the limited implementation experience I have with `cmake`.
I could not quickly find the name of the target directory for the `.exe` and had to synthesize it
right where you need it to change for a release configuration.

Let's look for another way that improve `CMakeLists.txt`, not erode it further.

For example, could you extract the target directory name to a variable and still have one `run_self_test`?
Now the problem is reduced to correctly initializing the target directory variable.
Maybe it already has a name. Good, use that name then.
Maybe the configuration (`Release`, `Debug`) is available in a `cmake` variable; use it in the initialization.
Maybe a command line parameter has to be created to specify our desired configuration.  Use it to decide. 

We have the time to get this right.  Let's do that and learn something together in the process.

rollbear:
I think the Debug/Release string is the value of `${CMAKE_BUILD_TYPE}`

PR #93, opened 2018-06-28
Improve support for C++17 and C++2a compiler modes

AndrewPaxie:
This fixes a `-Wdeprecated-declarations` warning when
`std::uncaught_exception()` is encountered.
Compiler: `g++-8` when compiling in `-std=c++17` or `-std=c++2a` mode.
Also triggered in `g++-latest` (g++ pre 9.0).

MSVC in Visual Studio 2017 15.7.4 generates C4996 when compiled with
`/std:c++17` or `/std:c++latest` flags.

Both these warnings are now resolved.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+	* Improve support for compiler modes:
+	  GCC and Clang: -std=c++17; MSVC: /std:c++17
+	  GCC and Clang: -std=c++2a; MSVC: /std:c++latest
+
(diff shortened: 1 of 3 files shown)


AndrewPaxie:
This may be merged independently of any changes required for Issue #94.

rollbear:
Note to self: A future C++20 mocking library should make better use of `std::uncaught_exceptions()`. It should check if the count has increased since before a call. I don't see this as easily doable with the current architecture.

PR #98, opened 2018-08-10
Automatic extern statements for compile-time reporters

rcdailey:
If you have multiple translation units, you can define the
`TROMPELOEIL_USER_DEFINED_COMPILE_TIME_REPORTER` preprocessor definition which causes the
`<trompeloeil.hpp>` header to automatically generate the following `extern` statement:

```Cpp
extern template struct trompeloeil::reporter<trompeloeil::specialized>;
```

This preprocessor definition can be added to your unit test code via your build system to make this
more transparent to your code. This definition provides the ability to define a compile-time
reporter without being required to insert code into multiple, existing translation units.

The old (legacy) requirement was that the `extern` statement above had to be explicitly declared in
every translation unit, which is tedious. However, this is still a functional alternative if you
can't define the preprocessor directive at the build system level for whatever reason.


rcdailey:
Note: I went with a fairly verbose macro name because later on I plan to add a `contrib`-type mechanism for automatic unit test framework integration. Right now documentation covers integration to many test frameworks, but this is code people have to put in their apps by hand each time. Eventually I will add more preprocessor definitions to automatically integrate to various test frameworks, which will override this macro. So this long name makes it generic enough later to use if you want to integrate a unit test framework that Trompeloeil doesn't know about.

rcdailey:
@rollbear Could you find the time to get me some feedback on this please? Thanks, much appreciated.

rollbear:
Sorry for the silence. I like the idea. I'm trying see if I can figure out something even neater.
Also, today was my first day on a new job, so my focus may be a bit elsewhere for a short while.

rcdailey:
No problem, just wanted to make sure you saw it. I am currently using this on a fork of mine for now so I'm not blocked. Congratulations on the new job, btw!

AndrewPaxie:
@rcdailey - in this pull request the `cookbook.md` file has one update to the
`extern template` requirement for multiple translation units, yet there are
more references in the sections for adapters to specific unit test frameworks.
Perhaps those references could become links back to the revised text you
wrote.

Would you raise an issue outlining the `contrib` mechanism you proposed
in the comment above?  It's a good idea and its merits should be stated
separately from this particular pull request.

As an aside, the Travis-CI build were having some trouble completing all the
jobs when you made the pull request.  I just pushed the retry button until they
succeeded.  We might want to keep an eye on the reliability of this service.

rcdailey:
I just realized that the `extern` cannot force explicit specialization of the template before that specialization is defined, which is the case in the 1 translation unit where we define the `send` function:

```cpp
#undef TROMPELOEIL_USER_DEFINED_COMPILE_TIME_REPORTER
#include <trompeloeil.hpp>

namespace trompeloeil
{
   template <>
   void reporter<specialized>::send(
      severity s,
      const char* file,
      unsigned long line,
      const char* msg)
   {
      std::ostringstream os;
      if (line) os << file << ':' << line << '\n';
      os << msg;
      auto failure = os.str();
      if (s == severity::fatal)
      {
         FAIL(failure);
      }
      else
      {
         CAPTURE(failure);
         CHECK(failure.empty());
      }
   }
}
```

I have to add the `#undef` to force the extern to not be defined in this case. Hopefully the implementation that @rollbear is going with will handle this case more elegantly.

To @AndrewPaxie's original question (sorry I'm late getting back to you), I probably won't make further updates to my PR since @rollbear said he wants to go with his own approach. I don't exactly know what he has in mind.

The contrib stuff I talked about I will work on once he's done, which is hopefully soon. And yes I will be opening separate issues about those to avoid mixing discussions here. Thanks.

rollbear:
Sorry for causing confusion. I currently don't have any better idea than yours, so if you have thoughts on how to improve it, then please go ahead and I'll eagerly await your new solution.

rcdailey:
Sure, I will put some more work into this then. Just to clarify, earlier you said:

> I'm trying see if I can figure out something even neater

What were you thinking? Do you have any good ideas/improvements to share? I'm curious what you were thinking at the time.

rollbear:
I was hoping to figure something out that didn't involve having to define a macro, but I've failed. I currently think it is impossible.

rcdailey:
Oh I was actually going to work on this some more later. I even kept a
browser tab open to remind me! I guess it can be a follow up commit. Sorry
for taking so long.

rollbear:
Feel free to continue. Just start from the current ´develop´ branch.

rcdailey:
@rollbear Sure, I can do that, but I'm more concerned with backward compatibility requirements this may have introduced that wouldn't have been there had I decided to rethink the solution. Hopefully not many people use it until then (not that I suspect it will change in such a significant way, but it's possible).

PR #100, opened 2018-08-16
Add FAQ about CMake usage

mlimber:


rollbear:
Neat. Thanks. There's yet another way too - use cmake `add_subdirectory` (one or two argument version.)
Would it read better with a small TOC for the 4 techniques, and add sub-headers for them? Something like:
```
* [add_subdirectory](#add_subdirectory)
* [local install](#local_install)
* [global install)(#global_install)
* [find_file)(#find_file)

### <A href="add_subdirectory"/>add_subdirectory()

... description
```

Also, please make the entry consistent with the others in the FAQ. The header for the `cmake` subsection should be H2, and the anrwer begin with an emphasized **A.**

```
## <A href="cmake"/>Q. How do I use this in a CMake project?

**A.** To use *Trompeloeil* in a project ...
```

mlimber:
I started to add the subheadings, but I backed it out because the local and global install share the "use in cmake" part, which breaks the separateness of the sections. I haven't used the `add_subdirectory()` method myself, so I just added it in passing at the bottom.

I have also addressed the headings. Not all of the existing ones were H2, so I fixed that as well.

I propose you accept my pull request and then make any further elaborations you feel are necessary. :-)

mlimber:
Any update on this?

rollbear:
I've just  been busier than I anticipated. Hopefully this weekend.

rollbear:
I think it is sufficient to just briefly mention the `add_subdirectory()` option. Anyone compelled to try it really should already be familiar enough with CMake to understand what it means.

PR #111, opened 2019-01-02
Move mock

rollbear:
I *really* want a 2nd pair of eyes on this one. By default mock objects are not movable, because allowing moves when there are live expectations can be really scary.

Allowing an explicit opt-in, with documented (scary) behaviour may be the way to go.

1. Does this actually really work safely?
2. Is the documentation effort on the right balance of correct, easy to understand, and writing on the nose?
3. Should this actually be done?

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -54,7 +54,7 @@
   - [`trompeloeil::stream_tracer`](#stream_tracer)
   - [`tropmeloeil::tracer`](#tracer_type)
   - [`trompeloeil::typed_matcher<T>`](#typed_matcher)
-  - [`trempoleil_movable_type`](#movable_type)
+  - [`trompeloeil_movable_type`](#movable_mock)
 - [Functions and Function Templates](#functions)
   - [`trompeloeil::expectation::is_satisfied()`](#is_satisfied)
   - [`trompeloeil::expectation::is_saturated()`](#is_saturated)
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -54,7 +54,6 @@
   - [`trompeloeil::stream_tracer`](#stream_tracer)
   - [`tropmeloeil::tracer`](#tracer_type)
   - [`trompeloeil::typed_matcher<T>`](#typed_matcher)
-  - [`trompeloeil_movable_type`](#movable_mock)
 - [Functions and Function Templates](#functions)
   - [`trompeloeil::expectation::is_satisfied()`](#is_satisfied)
   - [`trompeloeil::expectation::is_saturated()`](#is_saturated)
@@ -64,6 +63,8 @@
   - [`trompeloeil::print(std::ostream&, T const&)`](#print)
   - [`trompeloeil::set_reporter(...)`](#set_reporter)
   - [`trompeloeil::sequence::is_completed()`](#is_completed)
+- [Constants](#constants)  
+  - [`trompeloeil_movable_mock`](#movable_mock)
 
 ## <A name="notions"/>Notions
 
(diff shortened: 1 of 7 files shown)


AndrewPaxie:
Thanks for pulling this together.

I'll look this over, though I also need to do some research into just what kind of class results
from allowing these operations.  This may help me answer this question:
Is it meaningful to the design of a mock object framework to allow this?

Might take me a few days, but I will start with an extended set of build configurations and update
here with the results.

rollbear:
No hurry. If this involved minor details, I wouldn't have asked for a 2nd opinion, so take whatever time you need to understand the consequences of the change. I myself didn't see the need at all, otherwise support would've been there from day 1.

Also ping @toby-allsopp, you're one of those who have asked for this functionality. Your insights would be highly appreciated.

docs/reference.md
```diff
@@ -2122,14 +2123,75 @@ protected:
...
+[**`.SIDE_EFFECT()`**](reference.md/#SIDE_EFFECT),
+[**`.RETURN()`**](reference.md/#RETURN) and
+[**`.THROW()`**](reference.md/#THROW) and their `**LR_**` counter parts, will
+not themselves be moved. This means that if they refer to member variables
+in the mock objects, they will refer to dead store after the move. Also, keep
+in mind the lifetime of expectations. If you move a mock object that has
+active expectations, they should almost certainly be
+[**`NAMED_REQUIRE_CALL()`**](reference.md/#NAMED_REQUIRE_CALL),
+[**`NAMED_ALLOW_CALL()`**](reference.md/#NAMED_ALLOW_CALL) or
+[**`NAMED_FORBID_CALL()`**](reference.md/#NAMED_FORBID_CALL).
```
toby-allsopp: What do you mean by "the ... lambdas ... will not themselves be moved"? Aren't the lambdas part of the expectations?

What problem can occur when using `REQUIRE_CALL` instead of `NAMED_REQUIRE_CALL`? In the non-named case I guess that the expectation is owned by the mock object so won't it be transferred to the new mock object on move-construction?

docs/reference.md
```diff
@@ -2122,14 +2123,75 @@ protected:
...
+test(...)
+{
+  auto m = transfer(immobile{}); // compilation error
+  ...
+}
+test(...)
+{
+  movable m;
+  auto e = NAMED_REQUIRE_CALL(m, func(3));
+  auto mm = transfer(std::move(m)); // OK, e now belongs to mm.
```
toby-allsopp: I'm confused by this - is there more than one kind of ownership involved here? `e` is a `unique_ptr`, so _it_ is owning something, in some sense. Does the mock object (`m` or `mm`) also own the expectation?

docs/reference.md
rollbear: I will need to think about how to phrase this better. This is actually quite wrong... Thanks.

docs/reference.md
rollbear: Wrong word there. `e now refers to mm` would probably be better, or `A call to mm->func(3) will now satisfy e`, or something like that.

toby-allsopp:
I tried this out in the little proof-of-concept I've been playing with and it seems to work perfectly.

To make it easier to write expectations that refer to members of the mock object in a way that interacts more nicely with moveable mocks, would it be possible to introduce something like `_0` that refers to the mock object in a similar way to how `_1` etc refer to the parameters to the mock method?

The opt-in mechanism is vulnerable to a few kinds of errors that might be confusing for users, for example

```
  using trompeloeil_movable_mock = ::trompeloeil_movable_mock;
```

or

```
  static const bool trompeloeil_movable_mock = true;
```

My first instinct would be to reach for a specific base class to opt into this functionality but I'm not up to speed with the tradeoffs involved.

rollbear:
I'll definitely go with the `constexpr` member variable. Thanks.

I very much like the idea of `_0` (or some such means) to access the mock object. I figure it's not very hard in C++14, but it's not obvious to me how to do this in the C++11 API (unless it's OK to just pass a `void*` there, but I'd rather not.)

PR #117, opened 2019-01-29
Templatized base types now work with mock_interface

### LINKED ISSUE #116: Templatized base types do not work with mock_interface
So I do some trickery to forward template types through `mock_interface` in the middle of a hierarchy. Example:

```cpp
   template<typename TStateMachine>
   class MockState : public trompeloeil::mock_interface<Fsm::State<TStateMachine>>
   {
   public:
      IMPLEMENT_MOCK1(Enter);
      IMPLEMENT_MOCK1(Exit);
      IMPLEMENT_MOCK1(Update);
   };
```

And here is it being used:

```cpp
class TestStateMachine : public Fsm::StateMachine<TestStateMachine> {};
class TestState : public Testing::MockState<TestStateMachine> {};
```

`Fsm::State` here requires a template type, which I forward through my `MockState`. This code won't compile because the `typename` keyword is not used in the implement mock macros. Here is the code that fixes the problem:

```cpp

#define TROMPELOEIL_IMPLEMENT_MOCK_(num, name) \
  TROMPELOEIL_MAKE_MOCK_(name,,num, typename decltype(::trompeloeil::nonconst_member_signature(&trompeloeil_interface_name::name))::type,override,)
#define TROMPELOEIL_IMPLEMENT_CONST_MOCK_(num, name) \
  TROMPELOEIL_MAKE_MOCK_(name,const,num, typename decltype(::trompeloeil::const_member_signature(&trompeloeil_interface_name::name))::type,override,)
```

I will submit a PR for this issue.

rcdailey:
When using a template base class as the parameter to `mock_interface`, the `IMPLEMENT_MOCKn()`
macros fail to compile due to `typename` missing before the `decltype(..)::type` statement.

Fixes #116

### DIFF
--- a/include/trompeloeil.hpp
+++ b/include/trompeloeil.hpp
@@ -4224,9 +4224,9 @@ template <typename T>
   TROMPELOEIL_IMPLEMENT_CONST_MOCK_(15, name)
 
 #define TROMPELOEIL_IMPLEMENT_MOCK_(num, name) \
-  TROMPELOEIL_MAKE_MOCK_(name,,num, decltype(::trompeloeil::nonconst_member_signature(&trompeloeil_interface_name::name))::type,override,)
+  TROMPELOEIL_MAKE_MOCK_(name,,num, typename decltype(::trompeloeil::nonconst_member_signature(&trompeloeil_interface_name::name))::type,override,)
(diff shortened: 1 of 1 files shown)


rollbear:
Thank you. Would you mind adding a test case, to prevent the bug returning n the future?

Issue #118, opened 2019-02-17
"Integrating with unit test frame works" documentation could be improved


RedDwarf69:
Reading https://github.com/rollbear/trompeloeil/blob/master/docs/CookBook.md#unit_test_frameworks, specifically the "Compile time adapter" section, one would reach the conclusion that this code would compile:

```
#define TROMPELOEIL_USER_DEFINED_COMPILE_TIME_REPORTER
#include "trompeloeil.hpp"

namespace trompeloeil
{
  template <>
  void reporter<specialized>::send(
    severity s,
    const char* file,
    unsigned long line,
    const char* message)
  {
    return;
  }
}
```

But it fails to compile, complaining about "specialization after instantiation".

Also, if I am an user looking for how to integrate with Catch I'm unlikely to read the "Compile time adapter" section. I will directly go to https://github.com/rollbear/trompeloeil/blob/master/docs/CookBook.md#adapt_catch which fails to mention TROMPELOEIL_USER_DEFINED_COMPILE_TIME_REPORTER. The exact same problem exists for all the other sample adaptations.

rollbear:
Thanks. Yes, this is bad.

rcdailey:
I thought I mentioned this in another issue a year or two ago, but I can't find it. I remember throwing around the idea of a 'contrib' package of header files that each provide integration with a specific test framework. I think that would be a lot more useful than maintaining documentation that people essentially have to copy & paste from.

barneydellar:
I've only looked at the section for Catch (https://github.com/rollbear/trompeloeil/blob/master/docs/CookBook.md/#adapt_catch) , but it is wrong. 

I used this in a single translation unit instead: 
```
auto unused = trompeloeil::set_reporter([](
    trompeloeil::severity s,
    char const *file,
    unsigned long line,
    std::string const &msg
) {
    std::ostringstream os;
    if (line) os << file << ':' << line << '\n';
    os << msg;
    auto failure = os.str();
    if (s == trompeloeil::severity::fatal)
    {
        FAIL(failure);
    }
    else
    {
        CAPTURE(failure);
        CHECK(failure.empty());
    }
});
```
and it worked fine. I didn't need to extern anything. Updating the docs would definitely help newcomers!

I am using Visual Studio 2017 15.9.7, and Catch2 v2.7.0.

rollbear:
[...]

PR #119, opened 2019-02-20
Update CookBook for newer reporter

mattgodbolt:
Noticed when using v33.


rollbear:
Eh, what? There are problems in exactly this section of the documentation (see issue #118) but I don't understand where the `std::string` comes from. What problem have you seen?

rollbear:
The suggested changes has now landed in tagged release v34, so closing this PR.

Issue #121, opened 2019-03-15
clang 7.0.1 and TSAN warns about a data race in trompeloeil::get_lock


jktjkt:
Please [see the minimal test case](https://github.com/jktjkt/clang-tsan-static-mutex-trompeloeil). It seems that with all of the following options are met:

- using clang 6.0 or 7.0.1
- using libc++ v6 or v7
- TSAN (`-fsanitize=thread`) enabled
- optimizing at least `-O2`

...then [line **666**](https://github.com/rollbear/trompeloeil/blob/470b827518db19c43d7487a27f6e6ee9eae7dd38/include/trompeloeil.hpp#L666) causes two separate mutexes to be created. This quite likely has something to do with multiple translation units.

I cannot help but point out that it cannot be a mere coincidence that the offending line is line 666.

Here's what the thread sanitizer reports. Note that these two threads each lock a "different" mutex even though the actual mutex *should* be the same as far as I can tell.

```
==================
WARNING: ThreadSanitizer: data race (pid=4888)
  Write of size 8 at 0x7b0800000028 by thread T1 (mutexes: write M5):
    #0 trompeloeil::sequence_handler<1ul>::retire() <null> (doit+0x4b4871)
    #1 trompeloeil::call_matcher<void (), std::__1::tuple<> >::run_actions(std::__1::tuple<>&, trompeloeil::call_matcher_list<void ()>&) <null> (doit+0x4b31e3)
    #2 trompeloeil::return_of<void ()>::type trompeloeil::mock_func<false, void ()>(std::__1::integral_constant<bool, true>, trompeloeil::expectations<false, void ()>&, char const*, char const*) <null> (doit+0x4af37c)
    #3 void* std::__1::__thread_proxy<std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct> >, main::$_0> >(void*) <null> (doit+0x4aed3d)

  Previous read of size 8 at 0x7b0800000028 by main thread (mutexes: write M9):
    #0 check(trompeloeil::sequence const&) <null> (doit+0x4b4f74)
    #1 main <null> (doit+0x4aeb63)

  Location is heap block of size 24 at 0x7b0800000020 allocated by main thread:
[...]

PR #123, opened 2019-03-23
Statically allocate recursive_mutex to address issue #121

rollbear:
Would you please take a look and see if it solves the problem without causing too bad new ones.

@jktjkt does this work for you?

### DIFF
--- a/include/trompeloeil.hpp
+++ b/include/trompeloeil.hpp
@@ -654,23 +654,19 @@ namespace trompeloeil
 
   class specialized;
 
-  namespace {
-    inline
-    const std::shared_ptr<std::recursive_mutex>&
-    get_mutex_obj()
-    {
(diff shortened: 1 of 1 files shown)


AndrewPaxie:
Yes I will.  I've kept silent on this one since I discovered every I'd said before was based on
my many misunderstandings.

rollbear:
No worries, but it's good to get an extra pair of eyes on it. As far as I've understood, it needs to be a very exotic system for the leaked recursive_mutex to cause problems, so I'm not really concerned there.

rollbear:
Thank you. `aligned_storage´ is obviously what should be used. It's what it's there for.

I'm ignoring the `owner` issue for now.

clang-8 works fine.

AndrewPaxie:
Should be good to go after Coveralls and Travis finish.

PR #125, opened 2019-03-31
update doc - catch2 integration with runtime adapter

piponazo:
I noticed that the section "Integrating with unit test frame works" was out of date for the section "catch + runtime adapter". It is the first time I try to use trompeloeil, and I was getting the following compilation error when following the current version of the documentation:

```
/home/luis/programming/trompeloeilExample/tests/testMain.cpp:27:6: error: could not convert ‘<lambda closure object>main(int, char**)::<lambda(trompeloeil::severity, const char*, long unsigned int, const char*)>{}’ from ‘main(int, char**)::<lambda(trompeloeil::severity, const char*, long unsigned int, const char*)>’ to ‘trompeloeil::reporter_func {aka std::function<void(trompeloeil::severity, const char*, long unsigned int, const std::__cxx11::basic_string<char>&)>}’
     });
```
By changing the lambda signature to what it is specified in `trompeloeil.hpp:735` I could fix the issue and have my first mocking tests running :wink: 

Note that I changed the 'const-side' just to match with the alias defined in the trompeloeil header.

### DIFF
--- a/docs/CookBook.md
+++ b/docs/CookBook.md
@@ -194,9 +194,9 @@ Before running any tests, make sure to call:
 ```Cpp
   trompeloeil::set_reporter([](
     trompeloeil::severity s,
-    const char* file,
+    char const* file,
     unsigned long line,
-    const char* msg)
+    std::string const& msg)
   {
     std::ostringstream os;
     if (line) os << file << ':' << line << '\n';


piponazo:
By the way ... I just noticed that in #119 @mattgodbolt was touching part of the documentation that is very close to the lines I changed here. I did not have problems when using the specialization of the reporter, but only when trying to use the runtime adapter.

I was actually surprised to see that `reporter_func` and `reporter::send()` have different signatures. Is that a design decision?

rollbear:
You're not the first to have noticed. See issue #118. Documentation and header structure is updated on branch [`adapters`](https://github.com/rollbear/trompeloeil/tree/adapters), which is planned to go in to the next stable release, probably already on Monday. Please have a look and let me know if there are problems that needs to be addressed.

rollbear:
To answer your question about the different signatures. No, that's a historic mistake that I cannot remember any reason for.

piponazo:
I took a look to your `adapters` branch and it looks good! I am looking forward to increase the version number in my conanfile soon :wink: 

I'll close this PR then, seeing that you already took care about this issue.

PR #133, opened 2019-05-09
Honour CMAKE_INSTALL_LIBDIR

RedDwarf69:


rollbear:
Thank you. I really have to figure out why the Travis-CI clang7 builds fail have started failing recently, but that's another story.

PR #139, opened 2019-05-21
Disable clang-tidy warning

RedDwarf69:
Not sure what's your policy about this kind of stuff, adding support for "random" tools. But in case you are interested...

### DIFF
--- a/include/trompeloeil.hpp
+++ b/include/trompeloeil.hpp
@@ -4331,7 +4331,7 @@ template <typename T>
     return {nullptr, 0ul, nullptr};                                            \
   }                                                                            \
                                                                                \
-  mutable TROMPELOEIL_LINE_ID(expectation_list_t) TROMPELOEIL_LINE_ID(expectations){}
+  mutable TROMPELOEIL_LINE_ID(expectation_list_t) TROMPELOEIL_LINE_ID(expectations){} // NOLINT(misc-non-private-member-variables-in-classes)
 
 
 #define TROMPELOEIL_LPAREN (


rollbear:
Getting rid of warnings is good (Trompeloeil CI builds are with silly high compiler warning levels, just because.) I much prefer it if the warnings can be dealt with without suppressing them. In this case it's possible by making the macro define the member variable private, but that has other consequences. I may need to ruminate a bit on this. What do you think @AndrewPaxie?

RedDwarf69:
I played a bit with that.

Obviously if you just put a `private` before that then everything after the macro becomes unexpectedly private.
You could put a `private:` before and a `public:` after, but there is any warranty it was public before? `TROMPELOEIL_MAKE_MOCKX` doesn't seem to work if it's private... not sure if it could potentially be protected...
So yeah, if you have a proper fix it would be better. But I don't know enough about trompeloeil and would be afraid of breaking something, so I went with the suppression.

rollbear:
Therein lies the problem. I'm trying to guess what the chances are that existing test code would break if we'd mandate that mocked functions must be public. I'm guessing the chances are very low indeed. It'd need a slight bit of code restructuring though, since the macro must end at a position where you can add a semicolon without getting another warning.

AndrewPaxie:
I too would like to resolve warnings before suppressing them as a last resort.  The next round of Travis improvements will run GCC builds with the equivalent of `-WEverything` (from `clang`).

It feels like more than a year since I started an investigation into `clang-tidy` warning removal from this codebase.  To move this forward I would like a discussion in issue #140, starting with `clang-tidy` versions, command line flags, and configuration files used in production code.

AndrewPaxie:
I have a patch going through regression now:

- GCC and Clang (all supported configurations) - Pass
- Visual Studio configurations: will be tested in the next day or so.

Clang-tidy warning appears to be resolved, but I may be wrong.

All mock functions have to be declared public.  I have checked
all the examples in the documentation and they are declared
public.  We even have notes to this effect in `Cookbook.md`,
section "Mocking private and protected member functions".

With that in mind, I have determined the desired accessibility
of entities created in macro `TROMPELOEIL_MAKE_MOCK_`
and determined which are intended to private, which public.

It may have been simpler to move `struct tag_type_trompeloeil`
to the end of the macro and made it public, but this would have
broken the intended encapsulation.

Instead I opted to introduce a new type, `struct unused`, whose
definition conveniently requires and accepts the trailing semicolon
that follows the macro.

For the intrepid, here is the patch,

```text
$ git diff include/trompeloeil.hpp

diff --git a/include/trompeloeil.hpp b/include/trompeloeil.hpp
index 426fb57..7bdd002 100644
--- a/include/trompeloeil.hpp
+++ b/include/trompeloeil.hpp
@@ -4234,6 +4234,7 @@ template <typename T>
   TROMPELOEIL_MAKE_MOCK_(name,const,num, decltype(::trompeloeil::const_member_signature(&trompeloeil_interface_name::name))::type,override,)
 
 #define TROMPELOEIL_MAKE_MOCK_(name, constness, num, sig, spec, ...)           \
+  private:                                                                     \
   using TROMPELOEIL_LINE_ID(cardinality_match) =                               \
     std::integral_constant<bool, num == ::trompeloeil::param_list<sig>::size>; \
   static_assert(TROMPELOEIL_LINE_ID(cardinality_match)::value,                 \
@@ -4279,6 +4280,10 @@ template <typename T>
     }                                                                          \
   };                                                                           \
                                                                                \
+  public:                                                                      \
+  /* An unused type declaration, defined at end of macro. */                   \
+  struct TROMPELOEIL_LINE_ID(unused);                                          \
+                                                                               \
   TROMPELOEIL_LINE_ID(matcher_list_t)&                                         \
   trompeloeil_matcher_list(                                                    \
     TROMPELOEIL_LINE_ID(tag_type_trompeloeil)*)                                \
@@ -4306,7 +4311,9 @@ template <typename T>
     pmf_s_t const s_ptr = &T_ ## name::trompeloeil_self_ ## name;              \
     pmf_e_t const e_ptr = &T_ ## name::trompeloeil_tag_ ## name;               \
                                                                                \
-    ::trompeloeil::ignore(s_ptr, e_ptr);                                       \
+    TROMPELOEIL_LINE_ID(unused)* u_ptr;                                        \
+                                                                               \
+    ::trompeloeil::ignore(s_ptr, e_ptr, u_ptr);                                \
                                                                                \
     return ::trompeloeil::mock_func<trompeloeil_movable_mock, sig>(            \
                                     TROMPELOEIL_LINE_ID(cardinality_match){},  \
@@ -4331,7 +4338,13 @@ template <typename T>
     return {nullptr, 0ul, nullptr};                                            \
   }                                                                            \
                                                                                \
-  mutable TROMPELOEIL_LINE_ID(expectation_list_t) TROMPELOEIL_LINE_ID(expectations){}
+  private:                                                                     \
+  mutable                                                                      \
+  TROMPELOEIL_LINE_ID(expectation_list_t) TROMPELOEIL_LINE_ID(expectations){}; \
+                                                                               \
+  public:                                                                      \
+  /* An unused type definition to make trailing semicolon acceptable. */       \
+  struct TROMPELOEIL_LINE_ID(unused) {}
 
 
 #define TROMPELOEIL_LPAREN (
```
@RedDwarf69 : If you do try this and it resolves the warning, let us know.

rollbear:
You beat me to it. Thanks! I agree with your conclusion. In fact, your solution is identical to mine, save for the name and implementation of the dummy type ;-) (parenthetically, I'm guessing, but haven't benchmarked, that making the dummy type an alias instead of creating a new struct *may* compile a tiny bit faster, because it's not a new name being introduced.)

We should clarify the `public` requirement in the docs. As you say, it's already implied, and briefly mentioned, but it's not clearly stated as a requirement.

AndrewPaxie:
The `struct unused` was not my first choice.

A function declaration `void unused()` lead to `-Wunused-member-function`,
then after "using" it, a `-Wundefined-func-template`.

I was worried that `using unused = void` would lead to an unused alias warning.

I went ahead and implemented your suggestion to find out if my fears would
be realized.  They have proved unfounded:  no warning from GCC, Clang, or `clang-tidy`.

Out of time at the moment to try on Visual Studio, but here's the latest patch:

```text
git diff include/trompeloeil.hpp
diff --git a/include/trompeloeil.hpp b/include/trompeloeil.hpp
index 426fb57..dcebbb5 100644
--- a/include/trompeloeil.hpp
+++ b/include/trompeloeil.hpp
@@ -4234,6 +4234,7 @@ template <typename T>
   TROMPELOEIL_MAKE_MOCK_(name,const,num, decltype(::trompeloeil::const_member_signature(&trompeloeil_interface_name::name))::type,override,)
 
 #define TROMPELOEIL_MAKE_MOCK_(name, constness, num, sig, spec, ...)           \
+  private:                                                                     \
   using TROMPELOEIL_LINE_ID(cardinality_match) =                               \
     std::integral_constant<bool, num == ::trompeloeil::param_list<sig>::size>; \
   static_assert(TROMPELOEIL_LINE_ID(cardinality_match)::value,                 \
@@ -4279,6 +4280,7 @@ template <typename T>
     }                                                                          \
   };                                                                           \
                                                                                \
+  public:                                                                      \
   TROMPELOEIL_LINE_ID(matcher_list_t)&                                         \
   trompeloeil_matcher_list(                                                    \
     TROMPELOEIL_LINE_ID(tag_type_trompeloeil)*)                                \
@@ -4331,7 +4333,12 @@ template <typename T>
     return {nullptr, 0ul, nullptr};                                            \
   }                                                                            \
                                                                                \
-  mutable TROMPELOEIL_LINE_ID(expectation_list_t) TROMPELOEIL_LINE_ID(expectations){}
+  private:                                                                     \
+  mutable                                                                      \
+  TROMPELOEIL_LINE_ID(expectation_list_t) TROMPELOEIL_LINE_ID(expectations){}; \
+                                                                               \
+  public:                                                                      \
+  using TROMPELOEIL_LINE_ID(unused) = void
 
 
 #define TROMPELOEIL_LPAREN (
```

rollbear:
I just made a short comparison, 5 compilations of each alternative, using clang-8 only, and I can't tell any time difference from noise, and the produced binaries (full debug with all symbols) are identical, so I guess it's a matter of taste.

AndrewPaxie:
My taste this time is for an alias declaration.

Preparing a pull request to replace this one.

Omitted will be any wrapping of `[[maybe_unused]]` or `__attribute__((__unused__))`
to decorate the alias declaration.   I'd rather wait for the compilers to start complaining
before adding attributes such as this.

PR #132, opened 2019-05-03
Various documentation updates

rcdailey:
I have made a few various changes in this branch. I have separated each change into its own commit. However, to make the PR process easier, I have put them all up for review here. If any items are not acceptable or need more careful discussion, I'm happy to pull them out into separate PRs on request.

Summary of some changes I made:

* Ignore patterns were not sufficient, especially on Windows, so I made use of gitignore.io to generate some ignore patterns. The patterns specific to Trompeloeil were kept (they are at the bottom). The top portion can be replaced (upgraded) in the future as needed.
* A `.gitattributes` file was missing. I added it for proper EOL handling on Windows that doesn't depend on user's Git settings (repository now enforces EOL rules which will get you cleaner & more consistent PRs in the future).
* I added `.vscode` settings file that defines a few language-specific formatting rules for Markdown files. In the future, anyone using VS code to edit, format, and preview `*.md` file edits can benefit from this workspace configuration.
* I added some changes for #131 in the last commit.

PR #141, opened 2019-05-30
Contributions documentation page + vscode markdown settings

rcdailey:
Continuation of #132, these changes were slightly more controversial. I do find these changes to be very useful, at least they were for my own contributions. I'd like to see these incorporated into the project so others can benefit as well. Not everyone is required to use VS Code, but those that do, it would be great to see those benefits here.

Detailed changes:

* Add contributions documentation page    
* Added VS Code workspace settings    
    * Markdown language style settings (e.g. line length)
    * Various Markdown Lint rules and exclusions

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,27 @@
+{
+    "[markdown]": {
+        "editor.rulers": [
+            80
+        ]
+    },
+    "markdownlint.config": {
+        "line-length": {
+            "code_blocks": false,
+            "tables": false
+        },
+        "ul-style": {
+            "style": "asterisk"
+        },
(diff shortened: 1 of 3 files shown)


rcdailey:
@rollbear Would like to continue discussion on this, if you have the time please.

rollbear:
I like the idea of having a guideline, and I strongly like the idea of having some kind of correctness check added to the CI builds. But I cannot possibly endorse a recommendation to use a tool I personally do not use.

I don't want configuration files for tools in the repo, unless those configuration files are used by the CI. The reason is that I don't want to tie anyone in to any particular set of tools, and I cannot possibly check the correctness of those configurations for all tools that people are using.

PR #145, opened 2019-06-09
Suppress clang unneeded memfun

rollbear:
Suppressing the clang warning feels less bad than suppressing clang-tidy complaint about the work-around for the warning. This at least makes it clear that the intent is that the functions aren't meant to be implemented.

Curiously, there are several other functions that are unimplemented, and for which there are no warnings from clang. I've tried to recreate a simple example to file a bug report for clang, but I've failed. I have no idea of what makes these two functions special.


AndrewPaxie:
The only context where I could get this warning to be generated was
when a class using mocked functions was declared in an anonymous
namespace.  See `test/compiling_test_11.cpp` lines 45-81
(and pretending there is a close brace after the class definitions).

Perhaps this will help you create an example that generates the warning.

Issue #135, opened 2019-05-15
Travis-CI fails clang++7 ASAN builds


rollbear:
There are build errors like the below:

/usr/bin/ld.gold: error: /usr/lib/llvm-7/lib/clang/7.1.0/lib/linux/libclang_rt.asan-x86_64.a(asan_allocator.cc.o): invalid section group 124 refers to earlier section 58

These are not problems with Trompeloeil itself. We could drop ASAN checks for clang-7 builds, but it'd be better to fix it.

rollbear:
I've just disabled sanitizers on the clang-7 travis-ci builds. Green builds are important, and I think the likelyhood that something slips through this one, when the builds for practically all other compilers are sanitized, is very slim.

I still prefer a proper solution, whatever that may be, so leaving the issue open.

rollbear:
No longer relevant since travis is not used anymore

PR #146, opened 2019-06-11
GCC 9 and Clang 8 CI builds

offa:
Two new CI builds: GCC 9 and Clang 8.

I had to disable the sanitizer for the latter as it failed with the same error mentioned in #135.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -149,6 +149,21 @@ matrix:
         apt:
           sources:
             - ubuntu-toolchain-r-test
+            - llvm-toolchain-trusty-8
+          packages:
+            - g++-6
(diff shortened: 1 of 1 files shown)


offa:
I didn't touch any code, why did the coverage decrease!? :confused:

rollbear:
This is great. Thank you. I've seen that the coverage can fluctuate very slightly like that. I presume it's a rounding error somewhere. -0.002% is much less than one line of coverage difference, which is of course impossible when using a tool that only reports line coverage.

offa:
Is it possible to set a threshold? A big red cross doesn't look nice on a PR :grin: .

PR #147, opened 2019-06-20
Allow mocking of functions with template return type

Viatorus:
I hope it works as expected. :)


rollbear:
Nice. I hadn't thought about that.

Minor nitpick.
If you look at the diff for `trompeloeil.hpp`, you accidentally introduced some white space noise in the
line breaks for the `TROMPELOEIL_PARAMxx()` macros.

I think this PR should go towards the `develop` branch, since it's not really fixing a bug, it's adding a convenience.

Viatorus:
See #149

PR #160, opened 2019-09-11
Update CookBook.md

ytimenkov:
Minor fixes so code has higher chances to compile if copypasted.
Added `inline` since this function has to be declared in the header but it's also defined here which violates ODR.


rollbear:
Ops. Thanks.

Issue #157, opened 2019-09-11
Cannot REQUIRE_CALL to function taking non-const reference to non-copy type


jonnymwalker:
Here is a sample program that I think should compile, but doesn't:

```
#include "crpcut.hpp"                                 
#include "trompeloeil.hpp"                            
                                                      
int main(int argc, char* argv[])                      
{                                                     
    return crpcut::run(argc, argv);                   
}                                                     
                                                      
struct Widget                                         
{                                                     
    Widget() = default;                               
    ~Widget() = default;                              
                                                      
    Widget(const Widget&) = delete;                   
    Widget& operator=(const Widget&) = delete;        
                                                      
    Widget(Widget&&) = default;                       
    Widget& operator=(Widget&&) = default;            
};                                                    
                                                      
bool operator==(const Widget& a, const Widget& b)     
{                                                     
    return &a == &b;                                  
}                                                     
                                                      
struct Gadget                                         
{                                                     
    virtual void doSomeWork(Widget& w)                
    {                                                 
    }                                                 
};                                                    
                                                      
struct MockGadget : Gadget                            
{                                                     
    MAKE_MOCK1(doSomeWork, void(Widget& w), override);
};                                                    
                                                      
TESTSUITE(Demo)                                       
{                                                     
    TEST(gadget_works_on_widget)                      
    {                                                 
        Widget w;                                     
        auto pGadget = std::make_unique<MockGadget>();
        REQUIRE_CALL(*pGadget, doSomeWork(w));        
        pGadget->doSomeWork(w);                       
    }                                                 
}   
```         
g++ 5.5 outputs this as its first error message:                                         
```
In file included from demo.cpp:2:0:                                                                                               
demo.cpp: In member function ‘virtual void Demo::gadget_works_on_widget::test()’:                                                           
demo.cpp:44:57: error: invalid initialization of non-const reference of type ‘trompeloeil::param_list_t<void(Widget&), 0ul> {aka Widget&}’ f
rom an rvalue of type ‘Widget’                                        
         REQUIRE_CALL(*pGadget, doSomeWork(trompeloeil::_));
                                                         ^            
trompeloeil.hpp:141:40: note: in definition of macro ‘TROMPELOEIL_CONCAT_’                                                                  
 #define TROMPELOEIL_CONCAT_(x, y) x ## y
                                        ^                                                                                                   
trompeloeil.hpp:4502:50: note: in expansion of macro ‘TROMPELOEIL_CONCAT’                                                                   
[...]

PR #162, opened 2019-09-16
Simplify wildcard

rollbear:
Issue #157 highlighted a bug in the wildcard handling, and the code had a few `#ifdef`'s and comments regarding the abilities of different compilers. I believe this fix is not only simpler, but also correct, without needing compiler work arounds. I have not, however, been able to test on compilers older than gcc 4.8.5, other than a handful manual shots via compiler explorer.

Doc needs update, but please have a look and see if there are things missing, and also if some of the conditionally supported tests can be permanently enabled.


rollbear:
As far as I can tell, the tests enabled by `TROMPELOEIL_TEST_NEG_MATCHER_FAILURES` and `TROMPELOEIL_TEST_RVALUE_REFERENCE_FAILURES` can be enabled across the board now, but haven't used older gcc-4.8 than 4.8.5. Unfortunately `TROMPELOEIL_TEST_OVERLOAD_FAILURES` still seems to be a necessary filter.

rollbear:
I've hand tested all the `#ifdef`:d out tests with gcc-4.8.1 in compiler explorer, and all pass except "ANY can select overload on const lvalue reference type". So, unless I've made some copy-paste error (perfectly likely) it looks like gcc-4.8 support using c++11 just got better.

AndrewPaxie:
Ok, having a look now.

AndrewPaxie:
### Summary

Some configurations are failing: clang-6.0 and later with `-std=c++17` or `-std=c++2a`.
This occurs with a custom matcher defined - see `compiling_tests_14.cpp` line 3469.

Doesn't happen on Travis as `CMakeLists.txt` and `.travis.yml`
is only testing C\+\+11 or C\+\+14.

Sorry, but I can't offer solutions at this time.

### Detail

Unfortunately I'm getting these failed configurations

```text
clang++-latest -std=c++2a
clang++-latest -std=c++2a -stdlib=libc++
clang++-latest -std=c++17
clang++-latest -std=c++17 -stdlib=libc++

clang++-8 -std=c++2a
clang++-8 -std=c++2a -stdlib=libc++
clang++-8 -std=c++17
clang++-8 -std=c++17 -stdlib=libc++

clang++-7 -std=c++2a
clang++-7 -std=c++2a -stdlib=libc++
clang++-7 -std=c++17
clang++-7 -std=c++17 -stdlib=libc++

clang++-6.0 -std=c++2a
clang++-6.0 -std=c++2a -stdlib=libc++
clang++-6.0 -std=c++17
clang++-6.0 -std=c++17 -stdlib=libc++
```

Sample error message (start and end with middle elided):

```text
/usr/lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/type_traits:2160:41: fatal error: recursive template instantiation exceeded maximum depth of 1024
    auto declval() noexcept -> decltype(__declval<_Tp>(0))
                                        ^
./include/trompeloeil.hpp:656:39: note: while substituting explicitly-specified template arguments into function template 'declval'
  using invoke_result_type = decltype(std::declval<F&>()(std::declval<A>()...));
                                      ^
./include/trompeloeil.hpp:1009:26: note: in instantiation of template type alias 'invoke_result_type' requested here
              typename = invoke_result_type<Pred, V&&, T...>>
                         ^
./include/trompeloeil.hpp:1010:5: note: in instantiation of default argument for 'operator type-parameter-0-0 &&<trompeloeil::predicate_matcher<(lambda at test/compiling_tests_14.cpp:3438:3),
      (lambda at test/compiling_tests_14.cpp:3450:5), trompeloeil::duck_typed_matcher<(lambda at test/compiling_tests_14.cpp:3438:3), const char *, const char *>, const char *, const char *> >' required here
    operator V&&() const;
    ^~~~~~~~~~~~~~~~~~~~
./include/trompeloeil.hpp:656:58: note: while substituting deduced template arguments into function template 'operator type-parameter-0-0 &&' [with V = trompeloeil::predicate_matcher<(lambda at
      test/compiling_tests_14.cpp:3438:3), (lambda at test/compiling_tests_14.cpp:3450:5), trompeloeil::duck_typed_matcher<(lambda at test/compiling_tests_14.cpp:3438:3), const char *, const char *>, const char *,
      const char *>, $1 = (no value)]
  using invoke_result_type = decltype(std::declval<F&>()(std::declval<A>()...));
                                                         ^

// ... many thousands of lines ...

/usr/lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/tuple:588:19: note: while substituting prior template arguments into non-type template parameter [with _Dummy = void]
        constexpr tuple(const _Elements&... __elements)
                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
./include/trompeloeil.hpp:3475:7: note: while substituting deduced template arguments into function template 'tuple' [with _Dummy = (no value), $1 = (no value)]
    , val(std::forward<U>(u)...)
      ^
test/compiling_tests.hpp:275:3: note: in instantiation of function template specialization 'trompeloeil::call_matcher<void (std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >),
      std::tuple<trompeloeil::predicate_matcher<(lambda at test/compiling_tests_14.cpp:3438:3), (lambda at test/compiling_tests_14.cpp:3450:5), trompeloeil::duck_typed_matcher<(lambda at
      test/compiling_tests_14.cpp:3438:3), const char *, const char *>, const char *, const char *> > >::call_matcher<trompeloeil::predicate_matcher<(lambda at test/compiling_tests_14.cpp:3438:3),
      (lambda at test/compiling_tests_14.cpp:3450:5), trompeloeil::duck_typed_matcher<(lambda at test/compiling_tests_14.cpp:3438:3), const char *, const char *>, const char *, const char *> >' requested here
  MAKE_MOCK1(foo, void(std::string));
  ^
./include/trompeloeil.hpp:4770:35: note: expanded from macro 'MAKE_MOCK1'
#define MAKE_MOCK1                TROMPELOEIL_MAKE_MOCK1
                                  ^
./include/trompeloeil.hpp:4080:3: note: expanded from macro 'TROMPELOEIL_MAKE_MOCK1'
  TROMPELOEIL_MAKE_MOCK_(name,,1, __VA_ARGS__,,)
  ^
./include/trompeloeil.hpp:4253:15: note: expanded from macro 'TROMPELOEIL_MAKE_MOCK_'
          new matcher {                                                        \
              ^
test/compiling_tests_14.cpp:3469:23: note: in instantiation of function template specialization 'mock_c::trompeloeil_l_tag_type_trompeloeil_275::foo<trompeloeil::predicate_matcher<(lambda at test/compiling_tests_14.cpp:3438:3), (lambda at test/compiling_tests_14.cpp:3450:5), trompeloeil::duck_typed_matcher<(lambda at test/compiling_tests_14.cpp:3438:3), const char *, const char *>, const char *, const char *> >' requested here
    REQUIRE_CALL(obj, foo(is_clamped("b", "d")));
```

### Typical failing command line

```text
clang++-8 -std=c++2a -Weverything -Wno-padded -ftemplate-backtrace-limit=0 -Wno-c++98-compat-pedantic -Wno-weak-vtables -Wno-exit-time-destructors -Wno-global-constructors -Wno-extra-semi-stmt -I build/catch -I ./include -o compiling_tests test/compiling_tests.cpp test/compiling_tests_11.cpp test/compiling_tests_14.cpp
```

AndrewPaxie:
Are you able to give a reference section in the standard or cppreference.com that
describes the matching rules for these operators?
I tried "Overload Resolution",
https://en.cppreference.com/w/cpp/language/overload_resolution,
but I may be a off target.

rollbear:
http://eel.is/c++draft/basic.type.qualifier#4

A `volatile` member function can be called on a non-volatile object, but it is less preferred than a non-volatile member function, just as a `const` member function can be called for a non-const object, but only if no non-const member function matches the call. This resolves the ambiguity of the overloads, because the `volatile` function is only tried if the non-volatile function is not a candidate.

I'll look into the c++17 and c++2a issues.

rollbear:
Phiew. Reproducing the problem locally, and fixing it, was trivial. Making travis-ci run relevant builds required a long sequence of quite horrendous curses, but now we have (a few) C++17 and C++2a builds, both with g++ and clang++. More can, and should, be added, but I think this is sufficient to feel confident that this hasn't broken C++17 or C++2a. Again, I have done some manual verification of the previously disabled tests from compiler explorer, but there's always a risk I have accidentally missed something.

AndrewPaxie:
Running local regression tests.  Should be able to report outcome in next half day.

rollbear:
Looking forward to it. Thanks!

AndrewPaxie:
The results are in: all configurations pass.

There's a lot to like about this pull request:

- Better support for C++11 mode, principally in
  wildcard and duck type matching,
  and negating matchers.
- More configurations tested in Travis:
  - clang++-8 -std=c++14 -stdlib=c++
  - clang++-8 -std=c++17 -stdlib=c++
  - clang++-8 -std=c++2a -stdlib=c++
  - g++-8 -std=c++17
  - g++-8 -std=c++2a
  - g++-9 -std=c++17
  - g++-9 -std=c++2a
- Update to Catch2 2.9.2 for testing.

Code looks good.  I say merge and benefit.

rollbear:
Thanks. I need to update the docs before merging, but it feels good that it has passed your scrutiny. It's kind of a non-obvious change.

PR #165, opened 2019-10-08
Added hook so that successful tests can be reported correctly

ToreMartinHagen:
Catch2 has a framework to report number of assertions and list
which testcases it has ran. This does not work with REQUIRE_CALL
in trompeloeil.

Given a test in catch2 like this

SCENARIO("Foo test")
{
    WHEN("foo called"){
        THEN("call barfunc"){
            REQUIRE_CALL(bar, barfunc(5, ANY(int))).RETURN(4);
            foo.foo(5, 6);
        }
    }
}

will be reported as this:
  test cases: 1 | 1 passed
  assertions: - none -

If you run the test with the -s option to include  successful tests in ouput
you get the same output as above

With this fix REQUIRE_CALL is counted as asserts and are included in the
successful output like this:
  footest is a Catch v2.8.0 host application.
  Run with -? for options

  Scenario: Foo test
         When: foo called
         Then: call barfunc
  /scratch/onetechmain/qt/test/HwModules/gpio/FooTest.cpp:50

  /scratch/onetechmain/qt/test/TestHelpers/trompeloeil/include/catch2/trompeloeil.hpp:52: PASSED:
  with expansion:
    "bar.barfunc(5, ANY(int))" != 0

  All tests passed (1 assertion in 1 test case)

### DIFF
--- a/include/trompeloeil.hpp
+++ b/include/trompeloeil.hpp
@@ -3,6 +3,7 @@
  *
  * Copyright Björn Fahller 2014-2019
  * Copyright (C) 2017, 2018 Andrew Paxie
+ * Copyright Tore Martin Hagen 2019
  *
  *  Use, modification and distribution is subject to the
  *  Boost Software License, Version 1.0. (See accompanying
@@ -872,6 +873,14 @@ namespace trompeloeil
   }
 
   template <typename T>
+  void
+  send_ok_report(
+    std::string const &msg)
+  {
+    reporter<T>::sendOk(msg.c_str());
+  }
+
+  template <typename T>
   struct reporter
   {
     static
@@ -881,6 +890,12 @@ namespace trompeloeil
       char const *file,
       unsigned long line,
       char const *msg);
+
+    static
+    void
+    sendOk(
+      char const *msg);
+
   };
 
   template <typename T>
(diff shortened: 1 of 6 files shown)


include/trompeloeil.hpp
```diff
@@ -894,6 +908,13 @@ namespace trompeloeil
       reporter_obj()(s, file, line, msg);
     }
 
+  template <typename T>
+  void reporter<T>::
+    sendOk(char const *msg)
+    {
+        (void)msg;
+    }
```
rollbear: I think I prefer it if you never name the parameter, so you don't have to silence the warning.

```Cpp
sendOk(char const* /*msg*/)
{
}
```

include/trompeloeil.hpp
```diff
@@ -3002,6 +3023,19 @@ template <typename T>
...
+  void
+  report_match(
+    call_matcher_list <Sig> &matcher_list)
+  {
+    std::ostringstream os;
+    for (auto& m : matcher_list)
+    {
+        os << m.name;
+    }
+    send_ok_report<specialized>(os.str());
```
rollbear: I'm not sure I understand the meaning of this string.

include/trompeloeil.hpp
ToreMartinHagen: Will do

include/trompeloeil.hpp
ToreMartinHagen: Leftover when I was debugging compilation error. Will remove it.

ToreMartinHagen:
The Change in catch/trompeloeil was mistakenly not part of the commit.

Will make the change on the developer branch and the other comments you have.

ToreMartinHagen:
Updates done in pull request 166

PR #166, opened 2019-10-14
Added hook so that successful tests can be reported correctly

ToreMartinHagen:
Catch2 has a framework to report number of assertions and list
which testcases it has ran. This does not work with REQUIRE_CALL
in trompeloeil.

Given a test in catch2 like this

SCENARIO("Foo test")
{
    WHEN("foo called"){
        THEN("call barfunc"){
            REQUIRE_CALL(bar, barfunc(5, ANY(int))).RETURN(4);
            foo.foo(5, 6);
        }
    }
}

will be reported as this:
  test cases: 1 | 1 passed
  assertions: - none -

If you run the test with the -s option to include  successful tests in ouput
you get the same output as above

With this fix REQUIRE_CALL is counted as asserts and are included in the
successful output like this:
  footest is a Catch v2.8.0 host application.
  Run with -? for options

  Scenario: Foo test
         When: foo called
         Then: call barfunc
  /scratch/onetechmain/qt/test/HwModules/gpio/FooTest.cpp:50

  /scratch/onetechmain/qt/test/TestHelpers/trompeloeil/include/catch2/trompeloeil.hpp:52: PASSED:
  with expansion:
    "bar.barfunc(5, ANY(int))" != 0

  All tests passed (1 assertion in 1 test case)

### DIFF
--- a/docs/CookBook.md
+++ b/docs/CookBook.md
@@ -175,6 +175,28 @@ location. An example is an unexpected call to a
 are no expectations. In these cases `file` will be `""` string and
 `line` == 0.
 
+### Status OK reporting
+It is possible to make an adaption to the reporter that will be called if
+an expectation is met. This can be useful for correct counting and reporting
+from the testing framework.
+
+Provide an inline specialization of the
+`trompeloeil::reporter<trompeloeil::specialized>::sendOk()` function.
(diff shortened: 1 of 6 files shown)


rollbear:
Neat! I like this. Tusen takk.

PR #169, opened 2019-11-15
Enable "Status OK reporting" in doctest

RedDwarf69:
With DOCTEST_CONFIG_TREAT_CHAR_STAR_AS_STRING this prints

```
trompeloeil/include/doctest/trompeloeil.hpp:49: SUCCESS: REQUIRE_UNARY( trompeloeil_mock_calls_done_correctly ) is correct!
  values: REQUIRE_UNARY( "foo.func()" )
```

and without

```
trompeloeil/include/doctest/trompeloeil.hpp:51: SUCCESS: REQUIRE_NE( doctest::String(trompeloeil_mock_calls_done_correctly), "" ) is correct!
  values: REQUIRE_NE( foo.func(),  )
```

This is the best I could think of, but @onqtam may have a better idea.

### DIFF
--- a/include/doctest/trompeloeil.hpp
+++ b/include/doctest/trompeloeil.hpp
@@ -40,6 +40,17 @@ namespace trompeloeil
       DOCTEST_ADD_FAIL_CHECK_AT(f, line, msg);
     }
   }
+
+  template <>
+  inline void reporter<specialized>::sendOk(
+    const char* trompeloeil_mock_calls_done_correctly)
+  {
+#ifdef DOCTEST_CONFIG_TREAT_CHAR_STAR_AS_STRING
+    DOCTEST_REQUIRE_UNARY(trompeloeil_mock_calls_done_correctly);
+#else
+    DOCTEST_REQUIRE_NE(doctest::String(trompeloeil_mock_calls_done_correctly), "");
+#endif
+  }
 }
 
 


onqtam:
I'm actually not sure if ```DOCTEST_CONFIG_TREAT_CHAR_STAR_AS_STRING``` should have 
been introduced in the first place... But it's too late now.

I have no clue what you are trying to achieve here (I've never used mocking) but perhaps this page could be deemed relevant: https://github.com/onqtam/doctest/blob/master/doc/markdown/logging.md#messages-which-can-optionally-fail-test-cases

RedDwarf69:
trompeloeil already uses ADD_FAIL_CHECK_AT/ADD_FAIL_AT to report to doctest when an assertion fails.

This is about reporting to doctest when an assertion succeeds. Now that you mention it maybe `ADD_MESSAGE_AT` is what should be used here. But:

- sendOk doesn't have information about the file/line where the REQUIRE_CALL/ALLOW_CALL macro is. Should it, @rollbear @ToreMartinHagen?

- ADD_MESSAGE_AT only prints a message. We are not trying to say "Good morning" here, we are trying to say "an assert (you don't know about) passed". @onqtam could doctest add something like `ADD_PASS_AT`? The difference with ADD_MESSAGE_AT would be that it would only be printed if the test suite is executed with `--success`, and the output summary would count it in `[doctest] assertions:`.

onqtam:
Hmmm having something like ```ADD_SUCCESS_AT``` (or whatever it's called) does indeed make sense for completeness. Can you open an issue in doctest about it? It would again need to accept the file and line though...

ToreMartinHagen:
I don't see the need for file/line in the sendOk. I normally have nested test cases and with Catch2 you can get a nice printing of your tests. So running the Catch2 tests with a **-s** and some greping will give you printouts like this

```
Scenario: CompactorMotor test
      Given: Normal operation with 300 ms delay when changing direction
       When: runBackward called
       Then: run backward, start timer, start rotation sensing and state is
             runningBackward
       When: runForward called
       Then: stop runBackward, stop rotation sensing, start timer with
             minimumTimeWhenChangingDirection
       When: timer expires
       Then: run forward, start timer, start rotation sensing and state is
             runningForward
```
For me this has much higher value then the additional printout I get from all the assertions. 
```
test/TestHelpers/trompeloeil/include/catch2/trompeloeil.hpp:53: PASSED:
with expansion:
  "runForwardPin.setValue(START_MOTOR)" != 0

test/TestHelpers/trompeloeil/include/catch2/trompeloeil.hpp:53: PASSED:
with expansion:
  "higherLevelApp.onStateChanged(CompactorMotorStatus::runningForward)"
  !=
  0
```

ToreMartinHagen:
I don't see any problems in adding file/line to sendOk if that makes life better for others. Then they can just be ignored in Catch2.

RedDwarf69:
If and when doctest adds ADD_SUCCESS_AT it will make sense to use it. But it's unclear when this will happen and to support current versions of doctest we would still need the current version of this patch.
So, as per https://github.com/onqtam/doctest/issues/309#issuecomment-554792967, I would still request this to be pulled.

PR #175, opened 2019-12-31
Update Conan packaging and add to Travis

kaidokert:
Any interest in getting the Conan packaging updated ? 
Changes:
- Made the test package build again
- Updated it to latest conventions used in https://github.com/conan-io/conan-center-index ( maybe should be submitted there ? )
- Added minimal tweaks to .travis.yml to trigger test package builds

It would be easy to add automatic uploads to Bintray repos from here, too, but maybe a better approach is to actually PR the package into conan-center-index, where it gets auto published ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,11 @@
 language: cpp
 dist: xenial
 
+env:
+  global:
+    - PYENV_VERSION=3.7
+    - PATH=$PATH:~/Library/Python/3.7/bin HOMEBREW_NO_INSTALL_CLEANUP=1
+
 matrix:
   include:
     - os: linux
@@ -190,7 +195,7 @@ matrix:
             - libelf-dev
             - libdw-dev
       env:
-        - COMPILER="clang++-8" SANITIZE=false CXXFLAGS="-stdlib=libc++" CXX_STANDARD="17"
+        - COMPILER="clang++-8" SANITIZE=false CXXFLAGS="-stdlib=libc++" CXX_STANDARD="17" CONAN_COMPILER=clang CC_VERSION=8
 
     - os: linux
       addons:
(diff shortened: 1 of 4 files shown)


rollbear:
Nice! Thanks.

Would it be preferable for the travis builds to install conan via pip3, to get rid of the death of python-2 support messages?

Would it be preferable for the travis builds to
`conan profile update settings.compiler.libcxx=${CC_LIBCXX}` to get rid of compatibility warnings?
 
It is (for now at least) not a problem for me to upload the packages manually when I tag a release. It is a manual job, but one that only takes a minute. It is worth thinking about, though.

kaidokert:
Yep i'll take a look at both. 

Trouble with pip3 is that most of the Travis envs [don't have it installed by default](https://travis-ci.org/kaidokert/travis-explore/builds) even if Python3 is there, so will need to separately install and enable it.

kaidokert:
Updated. 

There doesn't appear to be a way to suppress this Conan warning, when not using profiles. Even when a default profile is generated like this 

```
      conan profile new --detect default
      conan profile update settings.compiler.libcxx=libstdc++11 default
```

Conan will still issue a warning when the subsequent create/package command isn't using a profile, see [code in detect_compiler_version here](https://github.com/conan-io/conan/blob/master/conans/client/conf/detect.py#L131). It's actually not looking at what's in the profile file at all in that case.

I explicitly didn't want to start creating profiles, they add a layer of obfuscation IMO.

Also, older Mac builds become super slow now when installing Python3 through Homebrew, about 22 minutes on Xcode 7.3 image. There isn't a very clean way to get around it, and brew caching isn't cleanly supported either. Alternative would be trying Pyenv installs and caching, if this is a significant concern

rollbear:
OK. Thanks for trying, though. It's not the end of the world, it's just that I automatically zoom in on warnings, wondering what the problem is. Having warnings I'm meant to ignore fills me with great discomfort (not to mention that another warning may appear that I miss, because I expect warnings.)

Accepting this one now. If you come up with something that is warning free, please submit another PR.

PR #176, opened 2020-01-06
Fix remaining issues with issue 173

AndrewPaxie:
#### trompeloeil.hpp

Modified `TROMPELOEIL_CONCAT` to work with old MSVC preprocessor.
This fixed a compile error in `TROMPELOEIL_REQUIRE_DESTRUCTION`
on MSVC compilers.  Not sure why Appveyor builds succeed.

Simplified expression for preprocessor guard around code
specific to GCC.

Narrowed scope of fallback implementation for `is_null_comparable`
to just MSVC 14.0 or earlier.

Removed an invocation of `TROMPELEOIL_REMOVE_PARENS` in call of
`TROMPELOEIL_PARAM_LIST` since the latter calls the former
internally.  Might be without effect but makes the code consistent
with other uses of the `TROMPELOEIL_PARAM_LIST` macro.

#### Unit tests

Tightened constraints on `null_constructible` to say explicitly
that comparison with `std::nullptr_t` is not supported.
I think this type is now an instance of the concept we wanted to model.
This change allows MSVC 14.0 to compile all test cases.

Enabled `null_constructible` test case as it was working with GCC 4.x
and now also passes with MSVC 14.0.

Modest improvements to test cases:

- More `INFO` on test case failure.
- Minor spelling fixes.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+	* Fix TROMPELOEIL_CONCAT macro issue to restore compiling with MSVC
+	  when traditional preprocessor is enabled.
+
+	* Fix issue 173 for GCC 4.x by performing testing to confirm
+	  correctness of existing code.
+
 v36 2019-04-01
 
 	* Status reporting of passed expectation calls as sucessful assertions in
--- a/include/trompeloeil.hpp
+++ b/include/trompeloeil.hpp
@@ -150,8 +150,17 @@
   TROMPELOEIL_IDENTITY(TROMPELOEIL_ARG16(__VA_ARGS__,                          \
                                          15,14,13,12,11,10,9,8,7,6,5,4,3,2,1,0))
 
+#if defined(_MSC_VER)
+
+#define TROMPELOEIL_CONCAT_(x, y, ...) x ## y __VA_ARGS__
+#define TROMPELOEIL_CONCAT(x, ...) TROMPELOEIL_CONCAT_(x, __VA_ARGS__)
+
+#else /* defined(_MSC_VER) */
+
 #define TROMPELOEIL_CONCAT_(x, ...) x ## __VA_ARGS__
 #define TROMPELOEIL_CONCAT(x, ...) TROMPELOEIL_CONCAT_(x, __VA_ARGS__)
+
+#endif /* !defined(_MSC_VER) */
 
 #define TROMPELOEIL_REMOVE_PAREN(...) TROMPELOEIL_CONCAT(TROMPELOEIL_CLEAR_,   \
   TROMPELOEIL_REMOVE_PAREN_INTERNAL __VA_ARGS__)
@@ -1075,7 +1084,7 @@ template <typename T>
   };
 
   struct indirect_null {
-#if defined(__GNUC__) && !defined(__clang__)
+#if TROMPELOEIL_GCC
     template <typename T, typename C, typename ... As>
     using memfunptr = T (C::*)(As...);
 
@@ -1085,7 +1094,7 @@ template <typename T>
     operator T C::*() const;
     template <typename T, typename C, typename ... As>
     operator memfunptr<T,C,As...>() const;
-#endif
+#endif /* TROMPELOEIL_GCC */
     operator std::nullptr_t() const;
   };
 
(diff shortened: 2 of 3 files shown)


test/compiling_tests.hpp
```diff
@@ -243,6 +243,7 @@ struct uncomparable_string {
 class null_constructible {
 public:
   null_constructible(int* p_)  : p(p_) {}
+  bool operator==(const std::nullptr_t&) const noexcept = delete;
```
rollbear: This is problematic. The idea is to ensure that an object is not implicitly constructed from a `nullptr` value. It is meant to emulate types like `std::string_view`. Explicitly deleting `operator==(std::nullptr_t)` makes this type an odd one, and not representative of the types that we want to avoid problems with.

AndrewPaxie: OK, I will try reverting this line, but it was necessary to resolve this issue with Visual Studio 2015. I suspect I am missing something in how `is_null_comparable` does its work.
An earlier attempt of mine made the `null_constructible` constructor `explicit`, before I
realized it was the wrong direction to take.

I also suspect it may be better in the short term for the test case not to be compiled,
so the GCC 4.x enabled build and `TROMPELOEIL_CONCAT` changes can merge sooner.
Consider the final state of this pull request a "part two of n" in that case, your original work
being "part one".

On the topic of the "problem types", `std::string_view` being one of them, I wondered while
working on this whether it would be worth creating a `compiling_tests_17.cpp` for test cases
specific to C++17.  A specific test with `std::string_view` might be one way to confirm
things haven't been broken (equivalently successfully fixed), rather that relying on
analogues of the types that are causing us problems.

I also wondered if it is worth encoding in `static_assert` statements the expected results
for Trompeloeil's type traits.  Early warning of compiler-specific differences might be useful.
We would also avoid testing them indirectly as we do now by relying on `print()` to test
`is_null_comparable`.

I further started to consider C++2a concepts as a means to model the expressions to have
hold for particular types, but it is too early in my learning about this language feature
to tell where it might lead.  Concept based checks might reasonably be placed in
`compiling_tests_20.cpp`.

Thoughts?

rollbear: Yes, yes, yes, yes.

Adding test suites for C++17 and C++20 is something we definitely should do. Not just ensuring that the C++11/14 constructs still builds. As for `std::string_view`, it was just the type that the problem was reported for. I'm sure there are lots of similar types written in C++11 code bases.

We should definitely do some static tests, maybe as static_asserts, for a lot of the metaprogramming hacks.

I'm less sure about adding C++17/20 constructs in the library. The reason is just that I hate the added complexity of having several interleaved code bases, separated by `#ifdef`. I much prefer to find solutions that works for all compilers and language versions. However, I can see two reasons for going there. The obvious one is to add new features to the library, that requires new functionality. That's a fairly harmless type of `#ifdef`. The other reason is if compilation time can be substantially reduced. For example, there's a fair amount of tag-dispatching going on, that could is C++17 be replaced with `if constexpr`. Likewise, in C++20, concepts could definitely be used. One such concept could be what this issue is about - null-comparable types. I feel that the perceived improvement for users must be quite noticeable for those changes, though, since it really adds a lot of complexity to already very complex code.

AndrewPaxie:
I have removed `operator==(const std::nullptr_t&)` from `null_constructible`
and disabled the failing test case in `compiling_tests_11.cpp` for
MSVC 14.0 only.  The `ChangeLog` is also more modest in what I claim
to have fixed.

Thank you for the very encouraging feedback on the direction my thoughts
have taken.  It's much appreciated, and I finally get why the
`null_constructible` class and its related types exist in the test suites.

I understand your caution with increasing the complexity of the
Trompeloeil header file.  Have you explored the `if constexpr` replacement
for tag dispatching, perhaps in a branch?  It might make a good starting
point for Trompeloeil Next Generation.

My thoughts on using concepts were to assist with the development of
test suites in the first instance i.e. nothing in the Trompeloeil header.
GCC has merged its concepts branch to HEAD, so there is at least one
configuration where experiments could take place.  I do want to follow
this up with a "paper" study of the concepts implicit in the current
Trompeloeil codebase.  This is why Trompeloeil is so rewarding: every
new feature of C++ has some possiblity to improve the library's
implementation.

I will move on to Visual Studio 2015, a separate pull request if I
find an implementation for `is_null_comparable`, but did want to check
a couple of static asserts with you.  Are these the expected results
for the following types:

```cpp
// From compiling_tests_11.cpp:
// Approx line numbers of test cases using these types in comemnts.

/* 3295 */
static_assert(
  trompeloeil::is_null_comparable<std::unique_ptr<int>>::value,
  "Failed: is_null_comparable<std::unique_ptr<int>>");

/* 3371 */
static_assert(
  trompeloeil::is_null_comparable<std::unique_ptr<int>&&>::value,
  "Failed: is_null_comparable<std::unique_ptr<int>&&>>");

/* 3448 */
static_assert(
  trompeloeil::is_null_comparable<std::unique_ptr<int> const&>::value,
  "Failed: is_null_comparable<std::unique_ptr<int> const&>");

/* 3963 */
static_assert(
  trompeloeil::is_null_comparable<null_comparable>::value,
  "Failed: is_null_comparable<null_comparable>");

/* 3997 */
static_assert(
  !trompeloeil::is_null_comparable<null_constructible>::value,
  "Failed: !is_null_comparable<null_constructible>");
```

PR #183, opened 2020-02-23
Rewrite iterator to avoid C++17 deprecation warnings

laudrup:
std::iterator has been deprecated in C++17 so rewrite iterator class
to provide the required typedefs instead.

This silences deprecation warnings from MSVC 2017.

### DIFF
--- a/include/trompeloeil.hpp
+++ b/include/trompeloeil.hpp
@@ -1498,9 +1498,13 @@ template <typename T>
 
   template <typename T, typename Disposer>
   class list<T, Disposer>::iterator
-    : public std::iterator<std::bidirectional_iterator_tag, T>
   {
     friend class list<T, Disposer>;
+    typedef std::bidirectional_iterator_tag iterator_category;
+    typedef T value_type;
+    typedef std::ptrdiff_t difference_type;
+    typedef T* pointer;
+    typedef T& reference;
   public:
     iterator()
     noexcept


laudrup:
This uses `typedef` instead of `using` since that seems to be required to work with MSVC 2015 according to this similar commit in [Google Protobuf:](https://github.com/protocolbuffers/protobuf/pull/5014/commits/2949ebd842e25c1bcb8b63b377a454a0887ea5a2)

AndrewPaxie:
Would you mind expanding a little to describe the command line flags
for MSVC 2017 which produce the "deprecation warnings" you have
observed?

laudrup:
> Would you mind expanding a little to describe the command line flags
> for MSVC 2017 which produce the "deprecation warnings" you have
> observed?

It seems to be enabled when building with [warning level **W3** and above](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4996?view=vs-2019).

No specific flag for enabling just that warning, but it does require setting the standard to C++17 (**/std:c++17**) or latest (**/std:c++latest**). The CI jobs for Windows currently targets C++14 so that would explain why it hasn't triggered there.

[Here](https://devblogs.microsoft.com/cppblog/c17-feature-removals-and-deprecations/)'s a blog post about it.

rollbear:
Thanks. Excellent. Could you add Appveyor jobs that builds with C++17, so that we don't accidentally reintroduce such problems again?

AndrewPaxie:
Strange that I never saw this when compiling with `/W4` and `/std:c++latest`
and Visual Studio Community 2019 16.4.5 building v140, v141, or v142 platform toolsets.
This does not comfort me. 
Any other warnings we have not resolved with MSVC?

laudrup:
@AndrewPaxie It is indeed a bit weird.

I experienced this while adding new tests to our codebase and for no obvious reason all of a sudden triggered this warning.

Strangely enough, this only happened when building with Clang on Windows which uses the same headers. I assumed that it was because of some strangeness in the compiler flags we use, but we are using the exact same compiler flags with MSVC as with Clang (ie. in clang-cl) so that couldn't explain it. I tried using the patch in this pull request and that fixed the issue.

Looking at the trompeloeil code it is quite clear that this is indeed deprecated and well documented that it would trigger this warning, so I'm fairly certain the fix is correct no matter what.

@rollbear I must admit I haven't actually seen this with the trompeloeil tests. I will try to see if I can reproduce this and add an Appveyor CI job if possible. Can't promise when I get around to it.

rollbear:
Regardless, you're right that `std::iterator<>` has been deprecated, so your PR is correct and desired. It'd still be good to have something in the CI builds. I'm also somewhat confused that we're not seeing warnings in the clang/libc++ builds with -std=c++17 or 2a.

laudrup:
@rollbear Thanks, good to hear and I agree that extending the CI builds would be a good idea.

The deprecation warning is only present in Microsofts STL implementation (it's a macro), so that's not that confusing. It's not related to clang, so what's confusing is that the warning triggered when building the code on Windows with clang and Microsofts STL but not Microsofts compiler.

include/trompeloeil.hpp
```diff
@@ -1498,9 +1498,13 @@ template <typename T>
 
   template <typename T, typename Disposer>
   class list<T, Disposer>::iterator
-    : public std::iterator<std::bidirectional_iterator_tag, T>
   {
     friend class list<T, Disposer>;
+    typedef std::bidirectional_iterator_tag iterator_category;
```
AndrewPaxie: Nitpick: In the original code these types are declared `public`.
Now they are declared `private`.  I suggest they remain `public`.

AndrewPaxie:
I have been able to configure my Visual Studio Community 2019 16.4.5 to compile with
`LLVM-clang-cl` and have
a) Reproduced the warning (`STL4015`).
b) Observed that the pull request code resolves the warning.

PR #187, opened 2020-04-07
Add runtime adapter for 'Status OK' reporting

### LINKED ISSUE #186: Runtime adapter for 'Status OK' reporting
We'd like to use the new "Status OK reporting" feature with Catch2 (really great addition, thanks!), but haven't found an easy way to enable it in our setup yet: We have a custom `main()` that calls `trompeloeil::set_reporter()`. As there is no equivalent to `set_reporter()` for the `sendOk()` override, we'd need to enable it in the tests itself - which, as they `#include <trompeloeil.hpp>` instead of `#include <catch2/trompeloeil.hpp>` , would apparently need modification to all of them (we also have no global test include file) :unamused: 

Would it be possible to add an `set_reporter(reporter_function, ok_reporter_function)` overload or similar? Given some pointers on how to implement this, I'd be happy to (try to) come up with a PR or help with testing.

moha-gh:
Provide a `set_reporter(reporter_function, ok_reporter_function)` overload to allow users to set a 'Status OK' reporter at runtime. Previously this was only possible at compile time by specializing `reporter::sendOk()`.

Fixes #186.

Please let me know whether I can / should add any tests for this.

### DIFF
--- a/docs/CookBook.md
+++ b/docs/CookBook.md
@@ -154,11 +154,15 @@ using reporter_func = std::function<void(
   char const *file,
   unsigned long line,
   std::string const &msg)>;
-reporter_func trompeloeil::set_reporter(reporter_func new_reporter)
+using ok_reporter_func = std::function<void(char const *msg)>;
+
+reporter_func trompeloeil::set_reporter(reporter_func new_reporter);
+reporter_func trompeloeil::set_reporter(
+  reporter_func new_reporter, ok_reporter_func new_ok_reporter)
 ```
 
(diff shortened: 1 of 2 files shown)


rollbear:
Thank you. I took the liberty of making a small change, the two function overload version of `set_reporter()` now returns a `pair<overload_func, ok_overload_func>`. I also made a few changes to the docs. Feel free to look into it, if you want to.

moha-gh:
Great, thanks. Looking forward to to the next release :)

PR #199, opened 2020-07-16
Support embedded, 32 bit plattforms

Viatorus:
Sometimes we run our unit tests on our target device (STM32).
In order to use trompeloeil, two changes are required:

1. Add an option set a custom recursive mutex type.

`std::recursive_mutex` is not available on bare-metal or RTOS. But it can provide a custom one.

2. Use `size_t` instead of `unsigned long long`.

`std::atomic<unsigned long long>` (used to keep track of the call count) is not available on 32-bit CPUs. I suggest to make this dependend on the size of `size_t` which is 4 bytes on 32-bit and 8 bytes on 64-bit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -427,6 +427,7 @@ before_script:
   - mkdir build && cd build
   - CXX=${COMPILER} cmake -DCXX_STANDARD="${CXX_STANDARD}" -DCMAKE_BUILD_TYPE=Debug -DTRAVIS_JOB_ID='${TRAVIS_JOB_ID}' -DSANITIZE=${SANITIZE} ..
   - make VERBOSE=1 self_test
+  - make VERBOSE=1 custom_recursive_mutex
 
 script:
   make run_self_test
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -203,6 +203,30 @@ if (MASTER_PROJECT AND CMAKE_BUILD_TYPE MATCHES Debug)
     )
   endif()
 
+  add_executable(
+    custom_recursive_mutex
+    EXCLUDE_FROM_ALL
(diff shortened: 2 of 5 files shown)


rollbear:
Weird that the CI checks were reported as passed when travis-ci failed 2 of the apple-clang tests (I believe the failures are spurious network problems at travis-ci.org and unrelated to your work.)

Thank you. Out of curiosity. why is it you need to run unit tests on target?

Also, can you please add documentation about `TROMPELOEIL_RECURSIVE_MUTEX`. It is currently only possible to find by reading the source.

AndrewPaxie:
The Travis failures for the two XCode builds boil down to `pip3` missing.
I saw these yesterday with issue 197/pull request 198.
Yes a separate issue, just adding my observations here.

rollbear:
Thanks @AndrewPaxie . But why on earth does the CI pipeline pass? We shouldn't hijack this PR for this problem, but it's an issue that needs to be fixed.

Viatorus:
Added the documentation.

> Out of curiosity. why is it you need to run unit tests on target?

Mostly the same reason this PR exists: Different OS, different integer sizes and there are interrupts. We want to make sure, everything is working correct as specified.

Viatorus:
Please wait with accepting this PR.

There is one unresolved issue. Since the custom mutex is not defined in the header <mutex>, the compiler can not find it. I will think about that. I will send a solution next Monday.

Viatorus:
Have a look on these changes. For embedded developer it is very common to define functions on there own if the compiler has no valid implementation for them (malloc/putc/...).

Viatorus:
@rollbear Any update on this?

rollbear:
Thanks for the reminder. I like this.

Issue #129, opened 2019-04-04
wildcard fails to match tuple of one type by value


rollbear:
Example:
```Cpp
struct S
{
    MAKE_MOCK1(func, void(std::tuple<int>));
};
using trompeloeil::_;
void q()
{
    S s;
    REQUIRE_CALL(s, func(_)); // Error, ambiguous
}
```

If the tuple takes two types, it works.

rollbear:
Ah, the compiler is confused if I want to match the tuple from the wildcard, or the tuple parameter. That's why it works if the tuple has 2 or more parameters, because then there is no ambiguity.

rollbear:
This can take a while, so if you think you know how to solve this, please check out branch `issue129`, enable the failing tests on
https://github.com/rollbear/trompeloeil/blob/issue129/test/compiling_tests_11.cpp#L1226-L1252 and
https://github.com/rollbear/trompeloeil/blob/issue129/test/compiling_tests_11.cpp#L1279-L1305

The problem is, at least partially, in the `wildcard` class on
https://github.com/rollbear/trompeloeil/blob/issue129/include/trompeloeil.hpp#L927-L979.

Send a PR when the test cases pass.

rollbear:
Included in tag v39

PR #203, opened 2020-08-21
Wildcard match tuple

rollbear:
This solves the old issue #129. I feel very uncomfortable with this fix, because I don't understand why it solves the problem. It provably does so, across several tests, which is obviously good, but I really want to understand why. I have asked, and received no answers. The solution was found by chance, not through analysis.

@AndrewPaxie ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -256,6 +256,19 @@ matrix:
           sources:
             - ubuntu-toolchain-r-test
           packages:
+            - g++-5
+            - libcurl4-openssl-dev
+            - libelf-dev
+            - libdw-dev
+      env:
+        - COMPILER="g++-5" SANITIZE=true CXX_STANDARD="17"
+
+    - os: linux
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+          packages:
             - g++-6
             - libcurl4-openssl-dev
             - libelf-dev
@@ -269,12 +282,38 @@ matrix:
           sources:
             - ubuntu-toolchain-r-test
           packages:
+            - g++-6
+            - libcurl4-openssl-dev
+            - libelf-dev
+            - libdw-dev
+      env:
+        - COMPILER="g++-6" SANITIZE=true CXX_STANDARD=17
+
+    - os: linux
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+          packages:
             - g++-7
             - libcurl4-openssl-dev
             - libelf-dev
             - libdw-dev
       env:
         - COMPILER="g++-7" SANITIZE=true
+
+    - os: linux
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+          packages:
+            - g++-7
+            - libcurl4-openssl-dev
+            - libelf-dev
+            - libdw-dev
+      env:
+        - COMPILER="g++-7" SANITIZE=true CXX_STANDARD=17
 
     - os: linux
       addons:
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -93,8 +93,7 @@ if (MASTER_PROJECT AND CMAKE_BUILD_TYPE MATCHES Debug)
              " -Wno-padded"
              " -Wno-weak-vtables"
              " -Wno-exit-time-destructors"
-             " -Wno-global-constructors"
-             " -Wconversion")
+             " -Wno-global-constructors")
       check_cxx_compiler_flag("-Wno-extra-semi-stmt" WARN_SEMI_STMT)
 
       if (WARN_SEMI_STMT)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -93,7 +93,8 @@ if (MASTER_PROJECT AND CMAKE_BUILD_TYPE MATCHES Debug)
              " -Wno-padded"
              " -Wno-weak-vtables"
              " -Wno-exit-time-destructors"
-             " -Wno-global-constructors")
+             " -Wno-global-constructors"
+             " -Wconversion")
       check_cxx_compiler_flag("-Wno-extra-semi-stmt" WARN_SEMI_STMT)
 
       if (WARN_SEMI_STMT)
@@ -107,7 +108,8 @@ if (MASTER_PROJECT AND CMAKE_BUILD_TYPE MATCHES Debug)
              "-Wall"
              " -Wextra"
              " -pedantic"
-             " -Wshadow")
+             " -Wshadow"
+             " -Wconversion")
     endif()
 
     set(WARN_FLAGS "${WARN_FLAGS} -Werror")
--- a/include/trompeloeil.hpp
+++ b/include/trompeloeil.hpp
@@ -1023,11 +1023,13 @@ namespace trompeloeil
 
   struct wildcard : public matcher
   {
-    template <typename T>
+    template <typename T,
+             detail::enable_if_t<!std::is_convertible<wildcard&, T>{}>* = nullptr>
     operator T&&()
     const;
 
-    template <typename T>
+    template <typename T,
+              detail::enable_if_t<!std::is_convertible<wildcard&, T>{}>* = nullptr>
     operator T&()
     volatile const; // less preferred than T&& above
 
--- a/include/trompeloeil.hpp
+++ b/include/trompeloeil.hpp
@@ -342,7 +342,7 @@
 
 #endif /* !(TROMPELOEIL_CPLUSPLUS == 201103L) */
 #if TROMPELOEIL_CPLUSPLUS > 201403L
-#  define TROMPELOEIL_INLINE_VAR inline
+#  define TROMPELOEIL_INLINE_VAR static inline
 #else
 #  define TROMPELOEIL_INLINE_VAR static
 #endif
--- a/include/trompeloeil.hpp
+++ b/include/trompeloeil.hpp
@@ -1048,7 +1048,7 @@ namespace trompeloeil
     }
   };
 
-  static constexpr wildcard const _{};
+  static wildcard _{};
 
 
 template <typename T>
--- a/include/trompeloeil.hpp
+++ b/include/trompeloeil.hpp
@@ -1024,7 +1024,7 @@ namespace trompeloeil
 
     template <typename T>
     operator T&()
-    const volatile; // less preferred than T&& above
+    volatile const; // less preferred than T&& above
 
     template <typename T>
     constexpr
@@ -1049,7 +1049,7 @@ namespace trompeloeil
   };
 
   static wildcard _{};
-
+  constexpr const void* wildcard_warning_avoider = &_;
 
 template <typename T>
   using matcher_access = decltype(static_cast<matcher*>(std::declval<typename std::add_pointer<T>::type>()));
--- a/include/trompeloeil.hpp
+++ b/include/trompeloeil.hpp
@@ -1023,13 +1023,19 @@ namespace trompeloeil
 
   struct wildcard : public matcher
   {
-    template <typename T,
-             detail::enable_if_t<!std::is_convertible<wildcard&, T>{}>* = nullptr>
+    template <typename T
+#if TROMPELOEIL_GCC && TROMPELOEIL_GCC_VERSION >= 50000
+              ,detail::enable_if_t<!std::is_convertible<wildcard&, T>{}>* = nullptr
+#endif
+              >
     operator T&&()
     const;
 
-    template <typename T,
-              detail::enable_if_t<!std::is_convertible<wildcard&, T>{}>* = nullptr>
+    template <typename T
+#if TROMPELOEIL_GCC && TROMPELOEIL_GCC_VERSION >= 50000
+              ,detail::enable_if_t<!std::is_convertible<wildcard&, T>{}>* = nullptr
+#endif
+              >
     operator T&()
     volatile const; // less preferred than T&& above
 
--- a/include/trompeloeil.hpp
+++ b/include/trompeloeil.hpp
@@ -341,6 +341,11 @@
   /**/
 
 #endif /* !(TROMPELOEIL_CPLUSPLUS == 201103L) */
+#if TROMPELOEIL_CPLUSPLUS > 201403L
+#  define TROMPELOEIL_INLINE_VAR inline
+#else
+#  define TROMPELOEIL_INLINE_VAR static
+#endif
 
 static constexpr bool trompeloeil_movable_mock = false;
 
@@ -1048,8 +1053,7 @@ namespace trompeloeil
     }
   };
 
-  static wildcard _{};
-  constexpr const void* wildcard_warning_avoider = &_;
+  TROMPELOEIL_INLINE_VAR wildcard _{};
 
 template <typename T>
   using matcher_access = decltype(static_cast<matcher*>(std::declval<typename std::add_pointer<T>::type>()));
--- a/include/trompeloeil.hpp
+++ b/include/trompeloeil.hpp
@@ -341,8 +341,8 @@
   /**/
 
 #endif /* !(TROMPELOEIL_CPLUSPLUS == 201103L) */
-#if TROMPELOEIL_CPLUSPLUS > 201403L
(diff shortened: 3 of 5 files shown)


AndrewPaxie:
Will offer what insight I can, along with a review.
Probably able to start in my weekend (next 24 hours).

rollbear:
Here's a miniature freestanding display. Why does removing `const` make it match all cases?

https://godbolt.org/z/s1d49T

rollbear:
I hate what this project is doing with my skills in handling obscure corners of C++ and compiler behaviour.

After adding substantially to Matt Godbolt's AWS bill, I believe I have a solution. It needs cleaning up.

```C++
#define TROMPELOEIL_REQUIRE_CALL_OBJ(obj, func, obj_s, func_s)                 \
  [&]{\
    _Pragma("GCC diagnostic push") \
    _Pragma("GCC diagnostic ignored \"-Wconversion\"") \
    \
    return ::trompeloeil::call_validator_t<decltype((obj).TROMPELOEIL_CONCAT(trompeloeil_self_, func))>{(obj)};\
    \
    _Pragma("GCC diagnostic pop") \
  }() + \
  [&]{\
    _Pragma("GCC diagnostic push") \
    _Pragma("GCC diagnostic ignored \"-Wconversion\"") \
    return ::trompeloeil::detail::conditional_t<false,                                \
                       decltype((obj).func),                                   \
                       decltype((obj).TROMPELOEIL_CONCAT(trompeloeil_tag_,func))>\
      {__FILE__, static_cast<unsigned long>(__LINE__), obj_s "." func_s}; \
    _Pragma("GCC diagnostic pop") \
    }() .func
```
I don't know why it needs to be in two lambdas, but I didn't get it to work with all logic in one lambda. I also haven't managed to get the _LAMBDA version of the macro to work.

rollbear:
So, got the lambda version to work too. Applying this diff to the sources at head of this branch, makes the self test program build warning free when using -Wconversion on g++-7 using C++17. It is tremendously ugly, though.

```diff
diff --git a/include/trompeloeil.hpp b/include/trompeloeil.hpp
index 24b8164..0db5567 100644
--- a/include/trompeloeil.hpp
+++ b/include/trompeloeil.hpp
@@ -4504,11 +4504,24 @@ template <typename T>
 
 
 #define TROMPELOEIL_REQUIRE_CALL_LAMBDA_OBJ(obj, func, obj_s, func_s)          \
-  ::trompeloeil::call_validator_t<trompeloeil_s_t>{(obj)} +                    \
-      ::trompeloeil::detail::conditional_t<false,                              \
+    [&]{\
+    _Pragma("GCC diagnostic push") \
+    _Pragma("GCC diagnostic ignored \"-Wconversion\"") \
+    \
+    return ::trompeloeil::call_validator_t<decltype((obj).TROMPELOEIL_CONCAT(trompeloeil_self_, func))>{(obj)};\
+    \
+    _Pragma("GCC diagnostic pop") \
+  }()\
+ +                    \
bf :-) confuciusornis ~/d/t/cmake-build-debug-gcc-7-17> make self_test 2>&1 | less
bf :-) confuciusornis ~/d/t/cmake-build-debug-gcc-7-17> make self_test 2>&1 | less
bf :-) confuciusornis ~/d/t/cmake-build-debug-gcc-7-17> make self_test 2>&1 | less
bf :-) confuciusornis ~/d/t/cmake-build-debug-gcc-7-17> make self_test 2>&1 | less
bf :-) confuciusornis ~/d/t/cmake-build-debug-gcc-7-17> git diff > /tmp/patch
bf :-) confuciusornis ~/d/t/cmake-build-debug-gcc-7-17> cat /tmp/patch 
diff --git a/include/trompeloeil.hpp b/include/trompeloeil.hpp
index 24b8164..0bc9062 100644
--- a/include/trompeloeil.hpp
+++ b/include/trompeloeil.hpp
@@ -4492,10 +4492,6 @@ template <typename T>
 #define TROMPELOEIL_REQUIRE_CALL_V_LAMBDA(obj, func, obj_s, func_s, ...)       \
   [&]                                                                          \
   {                                                                            \
-    using trompeloeil_s_t = decltype((obj)                                     \
-                              .TROMPELOEIL_CONCAT(trompeloeil_self_, func));   \
-    using trompeloeil_e_t = decltype((obj)                                     \
-                              .TROMPELOEIL_CONCAT(trompeloeil_tag_,func));     \
                                                                                \
     return TROMPELOEIL_REQUIRE_CALL_LAMBDA_OBJ(obj, func, obj_s, func_s)       \
       __VA_ARGS__                                                              \
@@ -4504,11 +4500,28 @@ template <typename T>
 
 
 #define TROMPELOEIL_REQUIRE_CALL_LAMBDA_OBJ(obj, func, obj_s, func_s)          \
-  ::trompeloeil::call_validator_t<trompeloeil_s_t>{(obj)} +                    \
-      ::trompeloeil::detail::conditional_t<false,                              \
+    [&]{\
+    _Pragma("GCC diagnostic push") \
+    _Pragma("GCC diagnostic ignored \"-Wconversion\"") \
+    \
+    return ::trompeloeil::call_validator_t<decltype((obj).TROMPELOEIL_CONCAT(trompeloeil_self_, func))>{(obj)};\
+    \
+    _Pragma("GCC diagnostic pop") \
+  }()\
+ +                    \
+    [&]{\
+    _Pragma("GCC diagnostic push") \
+    _Pragma("GCC diagnostic ignored \"-Wconversion\"") \
+    using trompeloeil_s_t = decltype((obj)                                     \
+                              .TROMPELOEIL_CONCAT(trompeloeil_self_, func));   \
+    using trompeloeil_e_t = decltype((obj)                                     \
+                              .TROMPELOEIL_CONCAT(trompeloeil_tag_,func));     \
+    return ::trompeloeil::detail::conditional_t<(bool)static_cast<std::remove_reference_t<trompeloeil_s_t>*>(nullptr), \
                                            decltype((obj).func),               \
                                            trompeloeil_e_t>                    \
-    {__FILE__, static_cast<unsigned long>(__LINE__), obj_s "." func_s}.func
+  {__FILE__, static_cast<unsigned long>(__LINE__), obj_s "." func_s};\
+    _Pragma("GCC diagnostic pop") \
+    }().func
 
 
 #define TROMPELOEIL_NAMED_REQUIRE_CALL_V(...)                                  \
@@ -4624,14 +4637,32 @@ template <typename T>
 #define TROMPELOEIL_NAMED_REQUIRE_CALL_(obj, func, obj_s, func_s)              \
   TROMPELOEIL_REQUIRE_CALL_OBJ(obj, func, obj_s, func_s)
 
-
+#if 0
 #define TROMPELOEIL_REQUIRE_CALL_OBJ(obj, func, obj_s, func_s)                 \
   ::trompeloeil::call_validator_t<decltype((obj).TROMPELOEIL_CONCAT(trompeloeil_self_, func))>{(obj)} + \
     ::trompeloeil::detail::conditional_t<false,                                \
                        decltype((obj).func),                                   \
                        decltype((obj).TROMPELOEIL_CONCAT(trompeloeil_tag_,func))>\
     {__FILE__, static_cast<unsigned long>(__LINE__), obj_s "." func_s}.func
-
+#endif
+#define TROMPELOEIL_REQUIRE_CALL_OBJ(obj, func, obj_s, func_s)                 \
+  [&]{\
+    _Pragma("GCC diagnostic push") \
+    _Pragma("GCC diagnostic ignored \"-Wconversion\"") \
+    \
+    return ::trompeloeil::call_validator_t<decltype((obj).TROMPELOEIL_CONCAT(trompeloeil_self_, func))>{(obj)};\
+    \
+    _Pragma("GCC diagnostic pop") \
+  }() + \
+  [&]{\
+    _Pragma("GCC diagnostic push") \
+    _Pragma("GCC diagnostic ignored \"-Wconversion\"") \
+    return ::trompeloeil::detail::conditional_t<false,                                \
+                       decltype((obj).func),                                   \
+                       decltype((obj).TROMPELOEIL_CONCAT(trompeloeil_tag_,func))>\
+      {__FILE__, static_cast<unsigned long>(__LINE__), obj_s "." func_s}; \
+    _Pragma("GCC diagnostic pop") \
+    }() .func
 
 #define TROMPELOEIL_ALLOW_CALL(obj, func)                                      \
   TROMPELOEIL_ALLOW_CALL_(obj, func, #obj, #func)
```

AndrewPaxie:
I don't know what "eyes emoji" is supposed to convey on this platform,
but it seemed appropriate, :smile: .

Your comment on `-Wconversion` was,

> It warns that the compiler does exactly what we want it to do.

This got me thinking that maybe it is one warning with two roles.
The first kind are those like `-Wnamespaces` and `-Wtemplates`,
a compiler-implemented convenience to identify certain code usage.
We don't usually want to enable these.

The second kind are those that indicate that what was written is
accepted by the language but may be not what you wanted.

I think on balance `-Wconversion` is of the second kind.

You have shown that with time and money the warning can be
suppressed.  This is a "local good" and is acceptable for use.

I am left wondering whether there is a more "global good"
to be found here.  The guiding question is "how do I write
code to avoid a conversion (and warning) in the cases where
a conversion is currently being added?"

I have some research to do!

rollbear:
The warning normally makes some sense, or at least has the potential to make sense. It is a sign that something may not be happening the way you think it does, if a class with conversion operators do not get the conversions called, but instead is copied. In this case, however, that is exactly what we want.

My work around, above, is problematic since it silences `-Wconversion` for all parameters when creating an expectation. What I'd like to do is to turn the warning off for the `wildcard` type, but I haven't been able to figure out a way to do that.

rollbear:
Ah, writing the above made me, maybe, come up with the better solution. Disable the conversion operators if wildcard *can* be implicitly converted to the desired type.

```diff
diff --git a/include/trompeloeil.hpp b/include/trompeloeil.hpp
index 24b8164..ce90f10 100644
--- a/include/trompeloeil.hpp
+++ b/include/trompeloeil.hpp
@@ -1023,11 +1023,11 @@ namespace trompeloeil
 
   struct wildcard : public matcher
   {
-    template <typename T>
+    template <typename T, std::enable_if_t<!std::is_convertible_v<wildcard&, T>>* = nullptr>
     operator T&&()
     const;
 
-    template <typename T>
+    template <typename T, std::enable_if_t<!std::is_convertible_v<wildcard&, T>>* = nullptr>
     operator T&()
     volatile const; // less preferred than T&& above
```

This builds without warnings with `-Wconversion` with g++-7 in C++17. More testing needed.

rollbear:
I had to limit this construction to only "recent" versions of gcc, but with that things seem to go well. I added `-Wconversion` flag to gcc and clang builds of `self_test`, and travis+appveyor passes. I don't have the fulll compiler/library/flag matrix, though, so perhaps there are some cases still missing.

AndrewPaxie:
Inspired solution!

Ran the full compiler/library/flag matrix locally and in general
the results are what we wanted.  Lingering, but not relevant:

- Still have the ICE in `g++-latest`
  (commit `68e605c93d57c17f07edd50f7e1c80f9216befd2`).
  This is not our present problem to solve.
- `clang++-9` has a local configuration issue that prevented it running.
  Also a different problem to solve.
- Matrix does not at present cover `g++-10`, `clang++-10`, `clang++-11`.
  I have some work to do!

A thought on `-Wconversion` in Clang builds.  I expect it to be enabled
with `-Weverything`, so there should be no reason to repeat it on the
command line.

Aside: A quick look at the Travis build environment suggests a
couple of `libc++` configurations could be immediately developed.
I will see which issue this belongs to, raise a new one if not there.

rollbear:
Right on the clang warning. I'll fix that later today.

I have tested successfully with g++ in C++11/14/17/20 with libstdc++, but I don't have a setup to try with libc++, and neither do I have a build of g++ from trunk. I guess I could try to reproduce on CompilerExplorer. The crash in an unreleased version doesn't bother me that much, but it must be reported so that it can be fixed before release.

Regarding the missing builds for Travis-CI. Can you add them on this branch (or maybe a branch forked from it)?

AndrewPaxie:
I'll add some "quick hits" when I can a little later this week,
but it will be quite incremental.  I will add on a branch forked from
this one.

rollbear:
With the aid of people more knowledgeable than me, I may have a reproduction of the g++-trunk ICE with libc++.

https://godbolt.org/z/8P5rj8

Reducing this to something I can report will take time and will be a job for another day.

AndrewPaxie:
Nice command line.

rollbear:
I don't know if this is the same ICE as the one you've seen, but I would've saved a lot of time if I had just started from nothing and built up, instead of starting from `trompeloeil.hpp` and removed stuff.

The smallest program I could come up with that causes sthe ICE is:

```Cpp
#include <tuple>
```
https://godbolt.org/z/PE5bGo

I'll write a gcc bug report. I don't know how insane the `<tuple>` header is, but a compiler crash is always a compiler bug, regardless of how crazy the code is.

rollbear:
gcc bug report: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=96819

AndrewPaxie:
Have completed local builds with a matrix extended to
`g++-10`, `clang++-9`, and `clang++-10`.  All pass.
(There is no `clang++-11` released at the time of writing.)

Will move forward with some simple infrastructure in Travis and cmake
to enable compilation with `libc++`, probably one `clang` compiler, then
one `gcc` compiler before going "all in" with the remaining
combinations.

What is your level of sensitivity to large increases
in the length of time Travis will take to complete the full matrix?
When complete, about twice the amount of existing work will be done.
With current resources, a 50 minute build will become something
like 100 minutes.

Your code to reproduce the ICE is delightful and speaks to the creation of
test cases that include just one of the standard headers, for all standard headers.
Looks like the standard library experts are already hard at work on your bug report.

The ICE I have may or may not be different.
I hope I can get as neat a program to reproduce it.
The Travis extensions will come first.

rollbear:
Long CI builds are problematic. They're already way too long for ideal. Another thing we could look into is to cache tools and kcov. I decided to use Cirrus-CI on another project, and it builds new images when you change your tools, and when you run again with the same configuration file, it saves the time for the download/install. Failed downloads are also the prime reason for failed CI builds, and that's just frustrating, and the risk will increase with a greater matrix.

One thing I think we can do, is to decide on which compilers/flags are low risk to not test. My current feeling on this is that we, for example, don't need to test C++11 support on the newer compilers (we already don't, but we should probably not add it). We don't need to test experimental language support (1x, 1y, 2a.) On the most modern compilers supporting c++20, it *may* be a fairly low risk to not test C++17.

AndrewPaxie:
I have made steady progress on supporting build configurations for
`g++` using `libc++` as the standard library.  I noticed `clang++-8`
builds have been using this library for the last year :smile: .

Actions remaining:

- Add one `g++` configuration to `.travis.yml` to use `libc++`.
- Rework command lines in `.travis.yml` for `clang++-8` to use
  new mechanism (specify `CXX_STDLIB` (new) on command line
  instead of co-opting `CXXFLAGS`).
- Test to ensure all this works in Travis.

rollbear:
I was wondering, do you think it's safe to merge this to master? I believe it almost certainly is. "Almost" is the weasel word,though.

AndrewPaxie:
I have high confidence in this pull request as it stands.

All `g++` and `clang++` configurations that I can test locally pass.
This is many more than currently enabled in Travis and at much
higher warning levels than enabled in `g++` on Travis at present.

AppVeyor passes the Visual Studio configurations.  

The remaining risk area to me is because I have not
tested the branch on my local installations of Visual Studio.
These installations are Visual Studio Community 2019,
versions 16.7.4 and 16.8.0 Preview 3.1.

Merging now removes the branch of a branch situation arising
for the `cmake/Travis` extensions, which could easily become
a separate pull request back to master.

AndrewPaxie:
I just tested this branch with Visual Studio Community 2019 (16.7.4)
with the following toolsets, all with ISO C++14 Standard selected:

- Visual Studio 2015 (v140)
- Visual Studio 2017 (v141)
- Visual Studio 2019 (v142)

Debug, Release for x64 and Win32 (x86) were tested.

All these configurations **passed** without error, and only the most minor
of warnings (LNK4075 in Release | Win32 - preexisting).

rollbear:
I think this is safe to say this one's solved for real now, so merging.

PR #205, opened 2020-09-23
Add missing `typename` keywords for trompeloeil_call_params_type_t

urdh:
This fixes an issue I was having when using mocks inside doctest's TEST_CASE_TEMPLATE macro.


rollbear:
Thank you!

PR #206, opened 2020-09-23
Allow using custom std::atomic and std::unique_lock implementations

urdh:
This is sort of in the same vein as #199. We're using trompe l'oeil on an embedded platform where the standard library doesn't provide `<atomic>` at all, and only provides an incomplete `<mutex>` from which `std::unique_lock` is missing.

These changes allow us to provide implementations of `std::atomic` and `std::unique_lock` based on the primitives provided by the RTOS, much in the same way #199 does for std::recursive_mutex.


rollbear:
Thank you!

PR #208, opened 2020-09-27
Gxx libcxx build

AndrewPaxie:
Extend Travis builds to include a job to use libc++ with g++.
Sanitizers are not presently supported, partly because of
an outdated version of cmake.


rollbear (commented):
On the phone now, so I haven't had a chance to test it, but can LIBCXX_PREFIX_PATH be defined on the cmake command line? If not, I think it should be. Other than that, I like the change. Thanks.

AndrewPaxie:
The `LIBCXX_PREFIX_PATH` may now be given on the `cmake` command line.
There is still a weird bug setting link flags with `set_target_properties()`.  It worked
with `target_link_options()` but that function is only in `cmake 3.13+` - so frustrating!

rollbear:
TIL about cmake `set(... CACHE ...)`. Nice!

PR #224, opened 2021-01-08
Add Alias target

Leon0402:
Hi,

just a quick PR to add an alias target, which has a couple of advantages. One of them is that it's clear that it's a cmake target:
target_link_libraries(myTarget PUBLIC trompeloeil). Here trompeloeil could the lib be a cmake target or a real lib name.
target_link_libraries(myTarget PUBLIC trompeloeil::trompeloeil) is unique though. It will look more consistent and in cases of errors the message will be more specific. It also avoids conflicts when having the library installed globally for instance.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,6 +14,7 @@ write_basic_package_version_file(
   COMPATIBILITY AnyNewerVersion)
 
 add_library(trompeloeil INTERFACE)
+add_library(trompeloeil::trompeloeil ALIAS trompeloeil)
 
 set(INCLUDE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/include)
 


rollbear:
Thank you. Please give me a few days to address the CI issue that shouldn't have failed your PR. I'll come back to you shortly. I see no reason to not accept this as is, but hang on a few days.

Leon0402:
Sure, no worries!

rollbear:
Ready to merge now. However, if you'd do me the favour of merging master into here, it'd be marvellous. Reason for asking is just to  see that the CI and coverage collection works correctly also for non-maintainers after the move to github actions and codecov.io.

rollbear:
Thank you, and apologies for the delay.

Leon0402:
No worries! I would have done the rebase as well today, just didn't have time before ;)

rollbear:
That's OK. If you want to, feel free to make a tiny PR with an update to the ChangeLog ;-)

PR #228, opened 2021-03-05
Fix documentation for `set_reporter` API in catch

mattgodbolt:
seems I had made this change several years ago and then sent the PR...to my own master branch instead of yours!


rollbear:
Thanks @mattgodbolt . It has indeed been a while, and that section has changed quite a bit since then. I went through the examples, and saw one case that had that error still, though, so I fixed it.

mattgodbolt:
Thanks!

PR #229, opened 2021-05-04
Upload coverage on mainline builds only

offa:
Upload coverage data on builds of this repository only. Trying on a fork will very likely fail all builds.


rollbear:
Ah, this makes sense. I did not think about it. Thank you!

PR #230, opened 2021-05-04
GCC 11 and Clang 12 builds

offa:
Adds GCC 11 and Clang 12 to the build matrix.

The `apt-get update` on gcc-11 is necessary at the moment, the package list seems outdated and doesn't include the latest gcc version yet.


rollbear:
Cool! Thanks. Much appreciated.

Issue #233, opened 2021-06-03
Printing family of types (class template)?


Manu343726:
Hi

I was wondering if there's a way to customize printing of a class template, say `std::optional<T>`. The docs say that `trompeloeil::print()` can be specialized to achieve this, but that approach only works with non-template types:

``` cpp
namespace trompeloeil
{
    template<>
    void print(std::ostream& os, const std::optional<int>& optional); // Works
    
    template<typename T>
    void print(std::ostream& os, const std::optional<T>& optional); // Doesn't work
}
```

> I'm not exactly a language lawyer but I would say the second one is overloading rather than specializing

rollbear:
Hmm, this seems to be an obvious over sight. I'll think about it. Thank you.

rollbear:
Alright, this is a design mistake that will probably become a breaking change to fix, unless I can come up with something very clever. The solution currently really requires specialization. An overload only works if it's visible before the call site (in other words, it works if you write the overload before `#include <trompeloeil.hpp>`, but that is clearly horrible).

Here's a work around for you, exposing some internals, that you can use in the mean while, but whatever fix will be made will probably break it.

https://godbolt.org/z/T1qvh6zEM

Thanks for reporting this. I'm surprised no one has reported it earlier.

Manu343726:
Maybe you can deprecate print() and use an ADL based trompeloeil_print(),
relegating the current default print() to do the usual ADL call.

That way the current print() specialization mechanism still works and users
can fully customize printing through usual ADL rules (very much like with
operator<<).

Anyway if I were you I would ask @vitaut, fmt has a long story of nicely
supporting customization points.

rollbear:
It seems this wasn't as bad as I feared. Please try branch https://github.com/rollbear/trompeloeil/tree/templated_print.

What you do here is to specialize a class `trompeloeil::printer<T>`. Like this:

```Cpp
namespace trompeloeil {
template <typename T>
struct printer<std::optional<T>> {
    static void print(std::ostream& os, const std::optional<T>& v)
    {
        // however you want to do this
    } 
};
}
```

The docs are updated on the branch. Please let me know if this solves your problem.

rollbear:
Merged the fix to master. Feel free to reopen if you disagree that this solves the problem.

rollbear:
Included in tag v41

PR #234, opened 2021-06-04
trompeloeil::printer<T> is the new customization point for printing (for #233)

rollbear:
Specializing `trompeloeil::print(ostream&, const T&)` is insufficient because function templates cannot be partially specialized, and overloads are resolved too early to be found. A class template `trompeloeil::printer<T>` is introduced as a preferred customization point, which does allow partial specializations.

Want to have a look @AndrewPaxie ? I'm reasonably confident about the implementation (although I'm sure it can be improved,) but less confident about the quality of the docs.


rollbear:
Never mind the clang-9 build failures. Those are caused by github-actions changing the images again (sigh). It's been taken are of on `master`.

PR #235, opened 2021-06-10
catch2: Fix build with CATCH_CONFIG_PREFIX_ALL

RedDwarf69:


rollbear:
I had not thought about that. Thank you!

PR #239, opened 2021-06-13
Issue 237: Disable -Wreserved-identifer warning from clang++-13

AndrewPaxie:


rollbear:
The comment is a bit misleading for the `-Wzero-as-null-pointer-constant`, since the compilers warn about tests that explicitly checks that *Trompeloeil* does the right thing when a user passes 0 as `nullptr`. Perhaps we should `#pragma` that away for those specific tests instead?

Also, we should have a way to remind us to re-enable these warnings once/if `Catch2` is fixed.

AndrewPaxie:
Yes, the scope of the comment I added was meant only to extend as far as the `-Wno-extra-semi-stmt`,
and I had nothing in particular to say about `-Wzero-as-null-pointer-constant`.  I would like to see the
effect of removing this warning, and, as you say, consider using `#pragma` in a restricted number of
contexts.

Catch3 contains a fix for the `-Wno-extra-semi-stmt`.  I can only hope the same happens for the
`-Wreserved-identifier` warning.  Perhaps the details can go into a "Migrate Trompeleoil tests to Catch3" issue.

PR #241, opened 2021-07-14
Verifies calls that are expected after an object dies

andy-saul-audinate:
Sequence calls are not retired when a deathwatched object is destroyed. This means that subsquent IN_SEQUENCE assertions on other objects are not honoured

Minimal bug repro: https://godbolt.org/z/Y7sWb163K

This PR resolves this issue

Hopefully the style is consistent with existing code. I couldn't find a clang-format or similar guidance. If the test name is too long, or if there are any other problems with this PR, please let me know :)


rollbear:
Very nice! Thank you.

Issue #242, opened 2021-08-05
CMake target namespace


offa:
If installed through CMake trompeloeil lacks a target `trompeloeil::trompeloeil`.

How about moving it to a namespace? That way it'll provide the mentioned target — unfortunately no plain `trompeloeil` anymore I guess though.

rollbear:
So weird. I was pretty sure this was added earlier this year. Commit 7bff1a432929bfada846a8d65554b77fe8ba7553 implies it, for sure.

Today is not the right day for me to dive into this, but I'll get back to you shortly. If you want to do git archaeology and find out where I absent minded that commit away, it'd be splendid. Thanks for reporting.

offa:
Seems alias targets don't get installed. I can provide a PR as adding the namespace isn't a big deal.

rollbear:
Please do. Thanks. Much appreciated!

rollbear:
Merged. Let's leave this open until the next release is tagged. Thanks for both the report and the fix.

rollbear:
Fix included in release v42

PR #243, opened 2021-08-05
CMake install namespace

offa:
Adds a namespace for the exported targets (#242).


rollbear:
Sometimes small changes makes big differences. Thanks!

Issue #259, opened 2021-10-10
trompeloeil gets installed with the project that includes it


ipazarbasi:
Hi!

I was working on install phase of my project, and found that trompeloeil is also installed along with my files. I don't know whether this is intentional.

I am adding an option to trompeloeil's CMakeLists.txt that disables installation when it isn't the top-level (master) project.

PR #260, opened 2021-10-10
Don't install trompeloeil if it isn't the top-level project

ipazarbasi:
Trompeloeil gets installed with the project that includes it, although
it's not used or required by the installed components. This patch
disables installation, if trompeloeil isn't the top-level project, and
fixes issue #259.


ipazarbasi:
Thank you for your feedback. I have update the FAQ in a new commit.

rollbear:
This is lovely. Thank you so much!

PR #266, opened 2021-12-07
ci: use setup-cpp to install the compilers

aminya:
This uses setup-cpp to install the C++ tools
https://github.com/aminya/setup-cpp/

https://github.com/aminya/trompeloeil/actions/runs/1551038387

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -958,6 +958,11 @@ jobs:
           ninja: true
           kcov: ${{ startsWith(matrix.config.os, 'ubuntu') }}
 
+      - name: Link xlocale.h for llvm 4.0 and older
+        if: matrix.config.libcxx
+        run: |
+          sudo ln -s /usr/include/locale.h /usr/include/xlocale.h
+
       - name: Verify compilation errors
         if: (!startsWith(matrix.config.os, 'windows'))
         shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -958,9 +958,12 @@ jobs:
           ninja: true
           kcov: ${{ startsWith(matrix.config.os, 'ubuntu') }}
 
-      - name: Link xlocale.h for llvm 4.0 and older
+      - name: Install dependencies for old LLVM and GCC
         if: matrix.config.libcxx
         run: |
+          # install libopcodes for collecting coverage
+          sudo apt install libbinutils
+          # link xlocale.h for llvm 4.0 and older
           sudo ln -s /usr/include/locale.h /usr/include/xlocale.h
 
       - name: Verify compilation errors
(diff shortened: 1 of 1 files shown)


aminya:
Does the libc++ installation script expect a certain llvm installation path? 
It builds fine, but the tests seems to fail for libC++
https://github.com/aminya/trompeloeil/runs/4441790990?check_suite_focus=true

rollbear:
Apologies for being a bit silent, but I will get back to you in a couple of days. I much appreciate the work you're doing. Thanks.

rollbear:
On the libc++ build/install-script. It's only there because not all versions have readily available downloads (as .deb packages.) If setup-cpp provides the right libc++ versions already, that step can be removed.

aminya:
> Apologies for being a bit silent, but I will get back to you in a couple of days. I much appreciate the work you're doing. Thanks.

You're welcome! 

> On the libc++ build/install-script. It's only there because not all versions have readily available downloads (as .deb packages.) If setup-cpp provides the right libc++ versions already, that step can be removed.

setup-cpp downloads the original llvm releases, which usually includes the libcxx. It does not use apt for this. So, I think we can try removing it

aminya:
The libcxx issue is fixed. It required also passing the CPPFLAGS and LDFLAGS to make.

aminya:
The only remaining thing is installing gcc-4.9, which for some reason is rejected by apt although its ppa is already added. How did you do that before? 🤔

aminya:
Yay! Everything is fixed now. 🎉

aminya:
The failures are due to missing coverage, which I think is unrelated to this PR. Right?

rollbear:
I'm not so sure that it is unrelated to the PR. If you look at the "Collect Coverage" step for gcc 4.8, it says:

```
kcov: error while loading shared libraries: libopcodes-2.34-system.so: cannot open shared object file: No such file or directory
```

Why is that library missing here?

aminya:
I updated kcov to v39 because it does not need to be built from the source this way. It might rely on some shared libraries that are missing in the old GCC installer. I need to try install missing packages and see if the issue is fixed.

rollbear:
Sorry for not paying much attention, but the push to finish things before Christmas was a bit intense.

I've looked into the current state, and tried to figure out how to move forward, but I haven't been that successful.

All builds that use `ubuntu-18.04` fail because `kcov` needs `binutils-2.34`, but the one installed is 2.30. I then tried to use `ubuntu-20.04` also for the older compilers, but that failed for completely different reasons. Now C++ standard headers cannot be found, so it looks like libstdc++ and libc++ are not available, or in paths that the compiler cannot find. I had hoped to reach a state where it all builds, but perhaps with some missing functionality, but unfortunately I have failed.

rollbear:
Of course, a different route forward, that may be easier, is to scrap `kcov` (it's not like I'm married to it) and use normal coverage builds instead. I have to read up on how to use it with the CI `codecov-action`, but I presume it's a normal enough use to be reasonably straight forward.

PR #274, opened 2022-08-03
Specify ARCH_INDEPENDENT when creating package version file as this i…

gregkalata:
…s a header-only library


rollbear:
Thanks. Never mind those clang-14 libc++ failures, those are obviously caused by something else that needs to be looked into. Please update the ChangeLog with this. Feel free to claim your rightful credit in there if you so wish.

gregkalata:
Updated, thanks! Great library btw.

rollbear:
Thank you for helping to make it better.

PR #297, opened 2023-04-06
On seq mismatch, report first in line and first required

rollbear:
@AndrewPaxie can you have a look and see if you understand why the regular expressions for sequence matching fail on the Windows builds? I assume it has something to do with line breaks, but I don't understand how. I've tried with explicit '\n' and the current shape with line breaks in the literal string source, hoping it'd become "\r\n".

### DIFF
--- a/include/trompeloeil.hpp
+++ b/include/trompeloeil.hpp
@@ -1981,6 +1981,11 @@ template <typename T>
       const
       noexcept;
 
+    bool
+    is_optional()
+      const
+      noexcept;
+
     void
     retire()
     noexcept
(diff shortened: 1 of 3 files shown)


rollbear:
I found a work-around, so no need, although you're of course still welcome to have a look.

I don't quite understand what's happening. A regex `"a\nb"` does not match. `"a\r?\nb"` does not either, which is really weird. `"a.*\nb"` **does** match, which is weirder still. Normally I would be very concerned by this, but since this is not the important bit I want to check for, I'm leaving it.

PR #301, opened 2023-04-26
Fix macro variadic arguments expansion on MSVC

gregkalata:


rollbear:
OK. Good, I think. Can you explain what problem this solves? Can you add a test that fails on MSVC without it?

rollbear:
Thank you! Much appreciated.

PR #302, opened 2023-05-05
fix example code in documentation

lbckmnn:
Some `using` keywords are missing in the code examples in the documentation.


rollbear:
Thank you!

Issue #313, opened 2023-09-29
Problematic traits for ``streamer`` primary template


DNKpp:
Hello,
with the arise of c++20 and its ranges namespace, providing views as parameters become more and more common. Unfortunatly ``trompeloeil`` doesn't fully support that, because the traits for the ``streamer`` primary template aren't precise enough.

Currently we have
```cpp
template <typename T,
          bool = is_output_streamable<T>::value,
          bool = is_collection<detail::remove_reference_t<T>>::value>
struct streamer
```

This is problematic in situations, where ``T`` provides ``begin`` and ``end`` members, but only for non-const access. This may and will happen, when you use some advanced views. Unfortuantly this results in a huge compile error.
Have a look at [std::views::owning](https://en.cppreference.com/w/cpp/ranges/owning_view#begin), which always offers a ``begin`` and ``end`` implementation, but conditionally enables the const overloads.

I think, the constraint can simply be fixed by simply adding const (at least in my project that works like a charm), as the streamer also uses a ``const T&``.

```cpp
template <typename T,
          bool = is_output_streamable<const T>::value,
          bool = is_collection<detail::remove_reference_t<const T>>::value>
struct streamer
```

EDIT: An alternative would be, to relax the ``streamer::print`` param to non-const, but this is probably not desired, as a ``streamer`` should not alter the streamed object in any way. So, I think it would be best, to stick with the ``const T&`` and strengthening the traits.

DNKpp:
So, any comments on this? Shall I compile a pull request or is there any information needed?

rollbear:
Sorry, I completely missed this. Thank you for pinging me. Please do make a PR, that would be very much appreciated.

rollbear:
Thank you so much for the fix. I'm leaving this open until a new release has been tagged.

rollbear:
Tagged v46 so closing now. Thanks again.

PR #317, opened 2023-10-06
Fix streamer for input-range like types

DNKpp:
As discussed #313

### DIFF
--- a/include/trompeloeil.hpp
+++ b/include/trompeloeil.hpp
@@ -1381,8 +1381,8 @@ template <typename T>
   using is_collection = is_detected<iterable, T>;
 
   template <typename T,
-            bool = is_output_streamable<T>::value,
-            bool = is_collection<detail::remove_reference_t<T>>::value>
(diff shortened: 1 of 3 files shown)


DNKpp:
Ah, lovely. Will try to fix that :)

rollbear:
If you want to use them, the docker containers with the various compilers are available at ghcr.io/rollbear/gcc and ghcr.io/rollbear/clang, e.g. `docker pull ghcr.io/rollbear/gcc:4.9`. See https://github.com/rollbear/ci_compilers/ for info on how to use them. Be careful with using them for other purposes - I change them randomly for whatever reason.

DNKpp:
Well, thanks. Hopefully it's working now. Never thought, that such a trivial code will be broken on c++11 conforming compilers.

rollbear:
Yeah, trying to get something to work with such a wide range of compilers of various ages can be quite challenging. Thank you so much for this fix.

PR #331, opened 2024-04-10
Matchers should receive non-const parameters when possible

DNKpp:
When working with matchers, I often stumble across a minor inconvenience when using matchers directly as a REQUIRE_CALL parameter.
Given the following situation:
```c++
struct Mock
{
    MAKE_MOCK1(foo, void(int));
};
```
``foo`` does not imply any restrictions on its param (it's a non-const value), but still any matchers receive it as const-ref => this custom matcher does not compile:
```cpp
const auto matcher = trompeloeil::make_matcher<trompeloeil::wildcard>(
	[](int& value)
	{
 return value == 42;
	},
	[](std::ostream&)
	{
	});
```
As its usually good style to provide read-only arguments as const-ref, this has some implications what we can actually check. In my case its the ``ranges::any_view`` from ``range-v3`` lib, which is actually unusable when constraint with const, as ``empty``, ``begin`` and ``end`` members are not const overloaded. 
Nevertheless there is sometimes the necessity to apply matchers on non-const refs, which currently is not directly supported. As the placeholders (``_1``, ...) do not promote to const, I can of course check with an ``WITH`` statement, but that seems just like a workaround to me.
The issue is, that you put each param into a tuple of references, but then pass the tuple around as const-ref, which then is supplied into ``std::get``, which actually does the const promotion on the elements. That's a pitty, but instead of creating a tuple of lvalue-refs, we can switch to a tuple of ``std::reference_wrapper`` and unwrap those where necessary.

### DIFF
--- a/include/trompeloeil/mock.hpp
+++ b/include/trompeloeil/mock.hpp
@@ -1000,6 +1000,15 @@ template <typename T>
   }
 
   template <typename T>
+  constexpr
+  bool
+  is_null(
+    std::reference_wrapper<T> t)
+  {
+    return is_null(t.get());
+  }
+
+  template <typename T>
   void
   print(
     std::ostream& os,
--- a/include/trompeloeil/mock.hpp
+++ b/include/trompeloeil/mock.hpp
@@ -1850,7 +1850,7 @@ template <typename T>
   template <typename R, typename ... T>
   struct call_params_type<R(T...)>
   {
-    using type = std::tuple<std::reference_wrapper<std::remove_reference_t<T>>...>;
+    using type = std::tuple<std::reference_wrapper<detail::remove_reference_t<T>>...>;
   };
 
   template <typename T>
--- a/include/trompeloeil/mock.hpp
+++ b/include/trompeloeil/mock.hpp
@@ -1152,6 +1152,20 @@ template <typename T>
       os << " matching _";
     }
   };
+
+  template <typename T>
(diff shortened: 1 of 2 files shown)


DNKpp:
As a side-note: Matching ``std::reference_wrapper`` params doesn't even work on current msvc, so this should break any existing code.
I tested this before my changes.
```cpp
struct Mock
{
	MAKE_MOCK1(foo, void(std::reference_wrapper<int>));
};

TEST_CASE_METHOD(
  Fixture,
  "C++11: Matchers work with std::reference_wrapper params.",
  "[C++11][C++14][matching]")
{
  const auto matcher = trompeloeil::make_matcher<trompeloeil::wildcard>(
	[](const std::reference_wrapper<int>& value)
	{
	  return value.get() == 42;
	},
	[](std::ostream&)
	{
	});

  {
    Mock u;
    REQUIRE_CALL_V(u, foo(matcher));
    int x = 42;
	u.foo(x);
  }
  REQUIRE(reports.empty());
}
```

DNKpp:
Sry, but which line are you talking about?^^

rollbear:
`u.func_lr(x);` is indented 6 spaces, the lines above it are indented 2 spaces.

DNKpp:
Oh, well. That was a tab :)
/done

rollbear:
Excellent. Thank you so much!

Issue #283, opened 2022-10-22
Non-constexpr TIMES


Swassie:
Hey

Currently the `TIMES` macro requires that its argument must be `constexpr`. This makes it not possible to use in a parametrized test or helper function where the number of calls depends on some parameter.

What is the reason behind this limitation, and can it be removed?

Thanks

rollbear:
The reason is that the limits are expressed in the type system, and are used for many compile time checks. It may be possible to make it a runtime limit, probably with somewhat worse compile time error discovery, but it would be a substantial change.

Swassie:
I think it would be a good change, especially since the workarounds (e.g. counter variables) aren't very good. The code would be cleaner and easier to understand if `TIMES` could be used in both cases.

rollbear:
PR is welcome. I think it may be possible, but I don't know how to just yet.

Swassie:
Sure, I can give it a try.

DNKpp:
I'm giving this a bump, as I would really love to see this to be possible.

rollbear:
This is now fixed on main after @DNKpp's PR was merged today. Thanks! I will close the issue once a release has been tagged.

rollbear:
Included in release v49

PR #333, opened 2024-04-14
add runtime ``TIMES``

DNKpp:
as discussed in this issue https://github.com/rollbear/trompeloeil/issues/283 it would be very helpful to have the TIMES action configurable with runtime arguments.
This PR aims to add that, without making too much changes to existing symbols (in fact, no changes are made at all).

As this is mainly a sketchup for the actual feature, I'm open for feedback regarding behavior and naming (maybe ``RT_TIMES is a more appropriate name).

### DIFF
--- a/compilation_errors/mixed_limits_with_rt_limits.cpp
+++ b/compilation_errors/mixed_limits_with_rt_limits.cpp
@@ -26,14 +26,14 @@ int main()
 #if (TROMPELOEIL_CPLUSPLUS == 201103L)
 
   REQUIRE_CALL_V(obj, f(),
-    .RT_TIMES(AT_LEAST(1))
+    .TIMES(AT_LEAST(1))
     .RT_TIMES(AT_MOST(3))
     .RETURN(0));
 
 #else /* (TROMPELOEIL_CPLUSPLUS == 201103L) */
 
   REQUIRE_CALL(obj, f())
-    .RT_TIMES(AT_LEAST(1))
+    .TIMES(AT_LEAST(1))
     .RT_TIMES(AT_MOST(3))
     .RETURN(0);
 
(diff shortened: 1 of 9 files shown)


DNKpp:
These detected memory leaks seem to be caused by the exception thrown from the ``dynamic_times`` constructor. But that's not the root cause of the issue, as ``trompeloeil`` surprisingly has a very poor resource managing (e.g. the matcher is "forwarded" via raw pointer).

rollbear:
From a very brief glance, this looks interesting. I'm away on a conference right now, but I'll have a closer look when I get back home again.

rollbear:
This is looking really neat. Thank you. Is there something you want to discuss, given that it's still marked [WIP]?

Otherwise, I think it's just a matter of adding docs and negative test(s) under 'compilation_errors' before I can merge it.

Naming can always be discussed, but I think DYN_TIMES is as good as anything I can come up with.

rollbear:
Ping! I know I haven't been very attentive lately, but I hope to be able to tag a release in the tear future, say during the weekend July 13-14th. If there are things you want from me to be able to merge this, then let me know. As mentioned before, I'm cool with the change and the names, but would appreciate docs and some more tests.

DNKpp:
Hey, sry for the delay. I'm currently very busy with my thesis, so I won't have much spare time for this during the next month or so.

I'm not sure, which tests I should add, as I mostly reuse existing components and already added a couple of tests. The negative tests are in the section beginning at 4835. Sure, we could explicitly add some constexpr tests, which will not compile with invalid bounds (because of the exception), but I don't think that would be of much value.

I would really like to rename it to ``RT_TIMES`` as this seems to me to be more in line with the rest of the framework (which already uses the 2 sign prefixes for other features).

rollbear:
Ok, no worries. There's no hurry.

DNKpp:
Hey, I finally got the time to invest some time into this feature.
I renamed it to ``RT_TIMES`` and added some compilation error files. Are they ok?

DNKpp:
So, I think I'm done. Added the necessary documentation and checked, that all anchors are working as expected. Don't know why codecov now started complaining.

rollbear:
Never mind codecov. It's being silly. I'll have a look. Thanks!

rollbear:
This is awesome. Thank you!

PR #334, opened 2024-04-14
Fix memory leak during expectation construction

DNKpp:
As discussed here https://github.com/rollbear/trompeloeil/pull/333 there exists a memory leak when creating expectations and a exception raises.

As its very hard to create a test for that, I use this PR and your the ci output as proof, that the issue is fixed. Hopefully its not too annoying. :)

See action run https://github.com/rollbear/trompeloeil/actions/runs/8680521953?pr=334 as a baseline, where I created a test to provoke the issue.


rollbear:
Wow. That's impressive. Thank you!

Issue #272, opened 2022-03-25
catch2/trompeloeil.hpp header requires bigger Catch2 include than is actually needed


johnbeard:
Catch2 v3 is no longer distributed as a single header. If you do this:

```
#include <catch2/catch_test_macros.hpp>
#include <catch2/trompeloeil.hpp>
```

Then the `CATCH_VERSION_MAJOR` will not be set, used here:

https://github.com/rollbear/trompeloeil/blob/6b2742870a4b08e87cf104e2e9e7904c3818275d/include/catch2/trompeloeil.hpp#L19

since that's in `#include <catch2/catch_version_macros.hpp>`. While this is included in Catch2's `catch2/catch_all.hpp`, the new split Catch2 headers apparently allow much faster compilation.

`catch2/trompeloeil.hpp` only actually needs the macros from `catch_test_macros.hpp` (specifically `FAIL`, `CAPTURE`, `CHECK` and `REQUIRE`).

It would be better, I think, if the check actually checked for what was used, perhaps like:

```
#if ! (defined(CATCH_CONFIG_PREFIX_ALL) && defined(CATCH_CHECK)) || (!defined(CATCH_CONFIG_PREFIX_ALL) && defined(CHECK))
#error "Catch2 macros must be defined before including <catch2/trompeloeil.hpp>"
#endif
```

(or it could do the include itself?)

rollbear:
Thanks. I'll look into it.

ingefredriksen:
Corrected fix:
```cpp
#if !((defined(CATCH_CONFIG_PREFIX_ALL) && defined(CATCH_CHECK)) || (!defined(CATCH_CONFIG_PREFIX_ALL) && defined(CHECK)))
#error "Catch2 macros must be defined before including <catch2/trompeloeil.hpp>"
#endif
```
Was missing parentheses.

PR #336, opened 2024-05-16
Update docs for Catch2 3.x integration

marktsuchida:
According to #272 and its fix #278, (only) `catch2/catch_test_macros.hpp` needs to be included from Catch2 3.x.
This updates the documentation accordingly.

- Clarify required header from Catch2 3.x
- Fix typo in header name (was: "catch/trompeloeil.hpp")


rollbear:
Awesome. Thank you!

PR #337, opened 2024-05-20
Allow specifying STDMETHODCALLTYPE to Mock methods

### LINKED ISSUE #329: Feature Request: Allow specifying calling convention attributes to Mock methods
A method could specify a calling convention attribute like `stdcall`, `cdecl`, `fastcall`, `thiscall`; supported by [GCC](https://gcc.gnu.org/onlinedocs/gcc/x86-Function-Attributes.html#x86-Function-Attributes-1), [Clang](https://clang.llvm.org/docs/AttributeReference.html#calling-conventions) and [MSVC](https://learn.microsoft.com/en-us/cpp/cpp/argument-passing-and-naming-conventions?view=msvc-170). which is part of the signature of the function. Currently we don't have a way to specify a calling convention in trompeloeil making it impossible to mock any function using these attributes.

[GMock](https://google.github.io/googletest/gmock_cheat_sheet.html#specifying-calling-conventions-for-mock-functions) does support a similar feature and allows to mock these kind of methods, so it would be nice to have it here too.

Looks like `TROMPELOEIL_MAKE_MOCK_` could be updated to accommodate another argument `calling_convention`
https://github.com/rollbear/trompeloeil/blob/45583abb9920e65d6f715378ddab690c42793a1f/include/trompeloeil/mock.hpp#L3524-L3526

And insert it in the generated method 
https://github.com/rollbear/trompeloeil/blob/45583abb9920e65d6f715378ddab690c42793a1f/include/trompeloeil/mock.hpp#L3586-L3589

Is this something you could consider?

cesenaLA:
Adds new `IMPLEMENT_STDMETHOD_MOCKn` and `MAKE_STDMETHOD_MOCKn`  to allow to mock COM interfaces declared using `STDMETHOD(name)` i.e. `virtual HRESULT __stdcall name (...).
`
As .idl/.tlb do not support cvref-qualified methods, so there's no need for all the permutations of `STDMETHODCALLTYPE` vs const/volatile/rvalue/etc.

The new macros are implemented by extending `TROMPELOEIL_MAKE_MOCK_` to take a calling convention and placing it in the correct location allowed by the microsoft compiler. This also means that the new macros only work on Windows.

It also add tests and documentation for the new macros.

Closes #329


rollbear:
I'm a bit confused by this. It solves a much more specific problem than the issue you originally opened. Is there a reason for limiting the support to only ´STDMETHOD`?

The change is good, BTW. Very good! Especially the docs and the extended test suite.

Issue #339, opened 2024-08-14
Mocking non-awaitable coroutines


chghehe:
Using Conan package `trompeloeil/48` and `gcc-14`.

The code below

```c++
#include <catch2/catch_all.hpp>
#include <catch2/trompeloeil.hpp>

#include <generator>

class Mock
{
public:
  MAKE_MOCK0(fibonacci,std::generator<int>());
};

TEST_CASE("fibonacci 7")
{
  Mock m;
  ALLOW_CALL(m, fibonacci())
    .CO_YIELD(0)
    .CO_YIELD(1)
    .CO_YIELD(1)
    .CO_YIELD(2)
    .CO_YIELD(3)
    .CO_YIELD(5)
    .CO_YIELD(8)
    .CO_RETURN()
    ;

  REQUIRE_THAT(m.fibonacci(), Catch::Matchers::RangeEquals(std::array{0, 1, 1, 2, 3, 5, 8}));
}
```

produces compile time error:

> trompeloeil/coro.hpp:54:56: error: ‘class std::generator<int>’ has no member named ‘await_resume’

Seems that non-awaitable coroutines are not supported (yet).

More than that, I cannot directly mock these functions (like `Mock::fibonacci()`) with `RETURN()` statement due to static assertions.

Since _at this moment_ `std::generator<>` is implemented in gcc-14 only, the following implementations might be used for reproduction purposes:
* `<experimental/generator>` with MSVC 192
* reference implementation from https://github.com/lewissbaker/generator
* any other implementation of synchronous co_yield generator

rollbear:
Thank you. I will have a look and see what I can do.

isliser:
Hi [rollbear](https://github.com/rollbear)
I hope you're doing well. I wanted to check in and see if you had a chance to look into

rollbear:
Thank you. Yes, I have, several times, but I'm afraid I have not been very successful. There's obviously something that I don't understand. It wouldn't surprise me if it's easy to do once I get it, but I'm not there yet, unfortunately.

chghehe:
I've noticed in tests, that provided `coro::generator<>` is awaitable and used via `co_await`, but generators are not awaitables and shall be used as range.

Also, `TIMES()` directive doesn't work properly with coroutines. The code below will succeed with less than 5 calls to `gen()`.

```c++
    co_mock m;
    REQUIRE_CALL(m, gen())
      .CO_YIELD(5)
      .CO_YIELD(8)
[...]

PR #344, opened 2025-03-25
Support range-like non-awaitable coroutines

chghehe:
Added support for coroutines like `std::generator` which cannot used with `co_await`, but might be iterated as a range.

#339

### DIFF
--- a/include/trompeloeil/coro.hpp
+++ b/include/trompeloeil/coro.hpp
@@ -14,6 +14,10 @@
 #ifndef TROMPELOEIL_CORO_HPP
 #define TROMPELOEIL_CORO_HPP
 
+#if __cplusplus < 202002L
+#  error "C++20 is required"
+#endif
+
 #if defined(__cpp_impl_coroutine)
 #  define TROMPELOEIL_COROUTINES_SUPPORTED 1
 #else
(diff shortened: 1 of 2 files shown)


rollbear:
Oh, it was that simple. I went in the completely wrong direction and made things very complicated for myself. Thank you.

I'll see if I can figure out what is causing problems with MSVC.

rollbear:
Grr. MSVC says 199711 for __cplusplus, even when compiling with /std:c++20. See https://godbolt.org/z/3MroGTe18

However, it does correctly set __cpp_impl_coroutine, and so do the other compilers, so your PR seems works fine if I remove the check for C++20. Was there a specific reason you added this, or can I merge with the check removed?

chghehe:
That is well known microsoft "feature" and might be solved by adding `/Zc:__cplusplus` compile option.
I use ranges in that header, that is why I added the check. If you think that is not necessary, feel free to remove it.

chghehe:
Also, I tried to use macro `__cpp_lib_ranges` but it is defined in `<version>` which is available only since c++20 :)

chghehe:
Probably, this check shall be refined with `_MSC_VER` and `_MSC_LANG` macros for MSVC... Or completely removed.

rollbear:
Just merged it. For some reason GitHub didn't recognize this as accepting your PR, so I'm closing it. The changes are in main now, though.

RedBeard0531:
@rollbear, would you mind cutting a release soon with this? If you could also squeeze in #351, that would be lovely 😄

PR #345, opened 2025-03-28
Fix feature detection of __VA_OPT__ for MSVC

### LINKED ISSUE #343: Traditional MSVC preprocessor does not have __VA_OPT__, even c++20 mode
I just encountered a weird problem where none of my mocks would compile after switching a project from Qt 5.15 to Qt6; when I tracked it down, the problem was that the new `MAKE_MOCK(name, auto (TArg) -> Treturn)` syntax wasn't getting signatures right, because `TROMPELOEIL_COUNT` started returning 0.

This project had been compiling with `-std:c++20` all along; the key thing that changed from Qt 5 to Qt6 is that CMake find_package machinery started adding the compiler flag `-Zc:__cplusplus`, so Qt can tell whether the *compiler* supports c++20 (regardless of the other components, like the preprocessor features).

Arguably MSVC is wrong here, of course, - if you don't support the new preprocessor features like `__VA_OPT__`, you don't really have complete c++20 support and shouldn't be claiming `__cplusplus >= 202002L`. But I think Microsoft's and Qt's choices are somewhat defensible - they don't do it by default, MS just offered [individually-tweakable `Zc:` conformance options](https://learn.microsoft.com/en-us/cpp/build/reference/zc-conformance?view=msvc-170) like `-Zc:__cplusplus` and `-Zc:preprocessor` if you want to shop ala carte. And Qt chose to force on as little as necessary to meet their needs. The annoying question is why `cl /std:c++20 /permissive-` still defaults to the traditional preprocessor... but that's just inertial.

But in any case, this combination makes the logic wrong at 

https://github.com/rollbear/trompeloeil/blob/eaeb89c1ce9d354b0ba7eb921fd5712cdbd78adf/include/trompeloeil/mock.hpp#L125-L136

if `_MSVC_TRADITIONAL` is defined, and set to 1, that indicates traditional MSVC preprocessor *regardless* of what `__cppstd` says, and `__VA_OPT__` is not available. So if you're bothering to cater to MSVC's traditional preprocessor at all, I think it this needs another clause:

```c
#elif defined(_MSVC_TRADITIONAL) && _MSVC_TRADITIONAL==1 
#  define TROMPELOEIL_HAS_VA_OPT 0
#  define TROMPELOEIL_MSVC_PREPROCESSOR 1
```

Or you could refactor it more, into something like

```c
#if defined(_MSVC_TRADITIONAL)
#  if _MSVC_TRADITIONAL==0 
#    if _MSC_VER >= 1940
#      define TROMPELOEIL_HAS_VA_OPT 1 
#    else 
#      define TROMPELOEIL_HAS_VA_OPT 0 
#    endif
#    define TROMPELOEIL_MSVC_PREPROCESSOR 0 
#  endif
#elif __cplusplus >= 202002L 
#  define TROMPELOEIL_HAS_VA_OPT 1 
#else 
#  define TROMPELOEIL_HAS_VA_OPT 0 
#endif
```

puetzk:
- Add VS2019 and VS2022 runs where `__cplusplus >= 202002L` is combined with `/Zc:preprocessor-` (and fail to preprocess correctly).

See https://github.com/puetzk/trompeloeil/actions/runs/14110751749 for an example where the new CI configurations (without the corresponding fix) fail to compile, because they try to use `__VA_OPT__` in a configuration where the preprocessor does not support it.

With the second commit (fixing the detection macro), all of the configurations now run, three of them with different behavior:
- "Windows MSVC 2019 C++20 preprocessor" should detect that it *does* have `__VA_OPT__`  (because it's >= 16.6, /`Zc:preprocessor` is on, and `/std:c++20` was set, making `_MSVC_TRADITIONAL=0` and `_MSVC_LANG=202002L`.

- "Windows MSVC 2019 C++20 __cppstd" and "Windows MSVC 2022 C++20 __cppstd" now see that they do *not* have `__VA_OPT__` (because it's using the old preprocessor, despite being `__cplusplus >= 202002L`

Fixes #343

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1676,6 +1676,13 @@ jobs:
             cxx_standard: 20,
           }
           - {
+            name: "Windows MSVC 2022 C++20 __cppstd",
+            os: windows-2022,
+            cc: "cl", cxx: "cl",
+            cxx_standard: 20,
+            cxx_flags: "/Zc:__cplusplus"
+          }
+          - {
             name: "Windows MSVC 2022 C++20 preprocessor",
             os: windows-2022,
             cc: "cl", cxx: "cl",
@@ -1728,6 +1735,13 @@ jobs:
             os: windows-2019,
             cc: "cl", cxx: "cl",
             cxx_standard: 20,
+          }
+          - {
+            name: "Windows MSVC 2019 C++20 __cppstd",
+            os: windows-2019,
+            cc: "cl", cxx: "cl",
+            cxx_standard: 20,
+            cxx_flags: "/Zc:__cplusplus"
           }
           - {
             name: "Windows MSVC 2019 C++20 preprocessor",
--- a/include/trompeloeil/mock.hpp
+++ b/include/trompeloeil/mock.hpp
@@ -122,13 +122,15 @@
 #define TROMPELOEIL_NOT_IMPLEMENTED(...) __VA_ARGS__
 #endif
 
-#if defined(_MSVC_TRADITIONAL) && _MSVC_TRADITIONAL==0
-#  if _MSC_VER >= 1940
-#    define TROMPELOEIL_HAS_VA_OPT 1
-#  else
-#    define TROMPELOEIL_HAS_VA_OPT 0
+#if defined(_MSVC_TRADITIONAL)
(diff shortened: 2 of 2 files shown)


rollbear:
Awesome! Thank you so much!

PR #365, opened 2026-07-20
Add support for free coroutine operators

### LINKED ISSUE #364: CO_RETURN does not work with non-member `operator co_await`
Per my (admittedly-weak) understanding of [cppreference's page on coroutines](https://en.cppreference.com/cpp/language/coroutines), an awaiter for `co_await expr` is found using the following logic:
1. Prefer `expr.operator co_await()` if it exists
2. Fall back to `operator co_await(static_cast<Awaitable&&>(expr))`
3. Fall back to `expr`, which is then assumed to have the usual await_suspend, await_ready, etc methods defined on it

However, trompeloeil only checks 1 & 3. This prevents it from working with e.g. WinRT coroutines like `winrt::IAsyncOperation`, which all rely on non-member `operator co_await`s: https://github.com/microsoft/cppwinrt/blob/a535d38f083e6a428b64ac47e3144653c9739bc5/strings/base_coroutine_foundation.h#L272-L296

I believe the fix would look something like this:
```diff
diff --git a/include/trompeloeil/coro.hpp b/include/trompeloeil/coro.hpp
index 8dbd881..2ca963f 100644
--- a/include/trompeloeil/coro.hpp
+++ b/include/trompeloeil/coro.hpp
@@ -51,6 +51,11 @@ namespace trompeloeil
         return type_wrapper<decltype(std::declval<T>().operator co_await().await_resume())>{};
       }
       else
+      if constexpr (requires(T coro) { operator co_await(std::declval<T>()); })
+      {
+        return type_wrapper<decltype(operator co_await(std::declval<T>()).await_resume())>{};
+      }
+      else
       if constexpr (requires(T coro){ coro.await_resume(); })
       {
         return type_wrapper<decltype(std::declval<T>().await_resume())>{};
```

If that seems reasonable, I can create a PR :).

\* AI disclaimer: I used AI to perform the initial root-cause of this issue while moving some projects off of MSVC's non-standard `/await` and onto standard C++20 coroutines. It also generated the first draft of the patch. However, this writeup is entirely my own (including performing the research on cppreference & finding C++/WinRT's co_await operators), and I also tightened the patch to properly reflect the `static_cast<Awaitable&&>` part that AI missed.

winstliu:
Fixes #364. I didn't end up adding any compilation_errors tests as this seems to be an instance where we are turning an existing failure case into something that now compiles - but let me know if you disagree.

(Tests written by AI, vetted by me before PR creation - though they're pretty much just carbon copies of the existing ones.)

### DIFF
--- a/include/trompeloeil/coro.hpp
+++ b/include/trompeloeil/coro.hpp
@@ -51,6 +51,11 @@ namespace trompeloeil
         return type_wrapper<decltype(std::declval<T>().operator co_await().await_resume())>{};
       }
       else
+      if constexpr (requires(T coro) { operator co_await(std::declval<T>()); })
+      {
+        return type_wrapper<decltype(operator co_await(std::declval<T>()).await_resume())>{};
+      }
+      else
       if constexpr (requires(T coro){ coro.await_resume(); })
       {
         return type_wrapper<decltype(std::declval<T>().await_resume())>{};
--- a/test/micro_coro.hpp
+++ b/test/micro_coro.hpp
@@ -342,6 +342,91 @@ namespace coro
   }
 
   template <typename>
+  struct co_await_task;
+
+  template <typename T>
+  struct co_await_promise : promise<T>
+  {
+    co_await_task<T>
+    get_return_object() noexcept;
+  };
+
+  template <typename T>
+  struct co_await_task
+  {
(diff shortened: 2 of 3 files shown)


rollbear:
Looking good. Thank you.

